PR #30, opened 2017-08-08
Add new PAIR_V1 protocol.

gdamore:
The PAIR_V1 protocol supports both raw and cooked modes, and has loop
prevention included.  In raw mode, it allows multiple connections to
be established (but requires the sender to identify which pipe by ID
it wishes to use), whereas in cooked mode there is just one.

We default to PAIR_V1 now.

My plan is to make all protocol versioning explicit in the future.


gdamore:
I will resubmit this later.

PR #79, opened 2017-09-18
Rename (moved) files into single src directory

mwpowellhtx:
Positions the code for easier integration with *SWIG* build steps.

I have not done anything else with the ``CMakeLists.txt`` yet, nor are there *make files* introduced, yet. That much is a work in progress.


gdamore:
Well, obviously this not done yet, as the code doesn't actually build.

I'm also thinking more about this, and now seeing what you've done (and having had time to sleep on it), have some questions.  The actual end user deliverables are currently in src/ already (src/core, etc.)  Some of the things that moved really don't belong in src/ because they are not part of the source code of the project (from my perspective at least).  For example, the contents of etc/ are ancillary tools to facilitate building, testing, editor configuration, etc.  These should have no interaction with SWIG.)

Some of the directories I can see room for debate on -- tests/ for example *are* source code, but they are not part of what I consider "deliverables" (meaning that while they are used to validate the build during build time, it is not the intention that they be packaged or installed).  Moving perf/ makes sense if you take the position that these sources are part of what we think *could* be installed or delivered.

I guess I'd imagine that the swig stuff should go into a directory called "swig" or something, but maybe I'm being a bit dense. 

What is the actual motivation behind having everything under a directory called src?  It doesn't make sense, IMO, to have a single src directory with no peers, since that just creates an arbitrary extra level of directory depth.  But again, maybe I'm misunderstanding the motivation here, and would like to better understand it if that is the case.

gdamore:
I've spent a little time looking for projects that integrate swig into their tree -- there arent many that use it for expressing their own API outward (more that use it for importing APIs, e.g. separate projects like  Python bindings to wxWidgets or somesuch).  This is one however http://download.osgeo.org/gdal/2.2.2/  -- and it doesn't have a collapsed src tree.  It does have a swig/ subdirectory.  Unfortunately it appears to be automake and GNUmake driven rather than cmake.

Still that might be a good example starting point.

mwpowellhtx:
I'm seeing the checks here, yes, I need to double check that then. Short of it is, no problem; I'll close this one out and resubmit.

It was an early draft to instill a clear demarcation between source code and the SWIG code. I see, *etc*, including shell, ASCII docs and sublime projects. I agree now that I look at it more closely, does not belong under *src*.

I'm not familiar with *dgal*, per se; although I am coming at it more or less using [*Google OR tools*](http://github.com/google/or-tools) as my model. I've seen it work there and subscribe to the C# packages, so I know this approach *can* work. Just a matter of *how* for a package such as *nng*.

At least in Google's case, they subscribe to SWIG among the "third party" components, so called. Then the SWIG source itself is provided under a *swig* directory same level as the *src* directory.

mwpowellhtx:
I double checked the CMakeLists are chaining properly. That seems to have done the trick. Stay tuned...

gdamore:
I'll have a look at the OR tools as well.

PR #80, opened 2017-09-18
Moved (renamed) source files to src directory and updated CMake files

mwpowellhtx:
Alright, hopefully this is a little saner. Corrected the CMake chaining which corrected the build results.

I noticed there is a *tools* directory during visual inspection which may be optionally included, but it is not committed. I put a *TODO* marker in the CMake files along these lines.

I also left the *etc* directory alone. Thanks for pointing that out. Although, looking at the CMake code a little closer, I'm not sure that these files aren't also "source"; i.e. for things like the *css* files, etc. Notwithstanding discussion around potentially *doxygen* adoption for HTML or other documentation, I'm not concerned about that right at the moment.


gdamore:
This looks better.  I do thing tools/ should probably be under src/ -- at least for tools that we package and deliver (such as nngcat).

The css file that is in /etc maybe should move ... right now we don't have any documentation being generated whatsoever.

I'm still on the fence about documentation systems -- I've looked at a number of them, and I've no particular love of doxygen.  There was clang-doc, which seems closer to what I want for generated documentation, but at the end of the day I think we want (need) to separate documentation of the internals from the documentation of the API *and* typical use cases.  Doxygen inputs are source code organized, and that isn't always the best way to provide consumer facing documentation.

We have to also provide documentation for things like transports, RFCs (protocol documentation), etc.  ASCIIDOC (and specifically asciidoctor) was the best compromise I could find at the time for this.  We may wind up also using doxygen style documentation for some things, but asciidoctor isn't going to be replaced by doxygen entirely.

mwpowellhtx:
Great, excellent. I was monitoring also, and figured it would pass this time when the *appveyor* passed.

I agree on the doc system. Not committed one way or another. Right now I just want to get a build system going and translating over to a SWIG output. I do like the idea of self-documentation sourcing the doc generation, however that gets generated, though. There might be an argument for a tests and examples source, however; examples including self-documentation generating docs. Anyway not a priority for me right now.

gdamore:
Looking more at this, and other CMake driven projects, I'm less convinced that this is worth the trouble it creates.  Are you still dependent on it/planning on it?

mwpowellhtx:
@gdamore I'm sorry, what? Depending/planning on what?

mwpowellhtx:
@gdamore Please explain. What started out as a fairly straightforward re-org, then you threw in a curve ball with CMake-driven projects. I'm not sure I follow on the one hand, much less the other. Porque?

gdamore:
What I mean by asking this, is that we have changed quite a bit already -- you're working in an out of tree repo (using nng as a subrepo), and using CMake and its support for external projects.  You're also using the .a now (or .LIB on windows) now, so I don't think you're very sensitive to things like the source tree organization; but I could be mistaken.

Given all this, I'm not sure if there is still a motivation present for reorganizing the tree.  I'm pretty happy with the organization the way it is (with the "main" nng library living in src/ and ancillary things like tests and performance tooling in other directories).

So what this amounts to is a question.  Do you still need to reorganize the tree?  What is the motivation here? If it comes down to nothing more than personal preference at this point, I'd prefer to just avoid the change, but if there are compelling reasons I'm open, I just want to understand them.

Hope that makes sense.

mwpowellhtx:
@gdamore You are correct. I wasn't sure what you meant, but this makes sense. Yes, thus far, from what I can determine, as long as I can the EP to works properly, that should be adequate to the task. No worries.

PR #111, opened 2017-10-10
fixes #110 Cancellation race in ZeroTier

### LINKED ISSUE #110: Cancellation race in ZeroTier
So it turns out that there is a cancellation in ZeroTier, because we don't check and guard against double completion.

We need to note going forward that all providers must take care to only "nni_aio_finish" a submitted AIO once - that means using locks to guard against double completion.

gdamore:


gdamore:
The only real test that should be interesting here is codacy.

PR #152, opened 2017-11-06
Add multiprotocol/multitransport stress test.

gdamore:


gdamore:
Codecov is unhappy because the multistress doesn't run every branch everytime it is run.  There is an element of randomness there; I'm ignoring it for now.

PR #162, opened 2017-12-12
Fix cut'n'paste typo

A-j-K:


A-j-K:
I have no idea how a small doc change can break the automation/build? Hmm.

gdamore:
Thanks.  The build failure was not your change -- I actually think it was a timing issue arising from the shared testing environment on macos.  Specifically, it looks like the "clock" related functionality time exceeded my expectations, causing the platform validation to fail. 

I've restarted the build -- I think that was a transient environmental issue.  I'll merge this shortly anyway, since we don't need that build given that this is just a doc fixup.

Issue #35, opened 2017-08-08
Add kqueue support for Darwin/BSD


gdamore:
We would like to have kqueue based polling for improved scalability.

liamstask:
Hi - I began taking a look at this since it looked like a reasonably well isolated task, and had a couple questions about your vision for the polling system.

Currently, the `poll` based poller is assumed for all posix platforms - it's enabled in posix_config.h and used throughout the posix platform code.

Since some posix platforms will want to use different polling implementations (kqueue, epoll, etc), do you imagine creating a polling api that would be implemented for those backends? Maybe you have another approach in mind?

Let me know and I can either try to assist or just stay out of the way until it's clearer how to proceed 😄

gdamore:
Sorry I didn't respond earlier.  Thanks for your interest in contributing to nng!

So I think you can look at the posix_pollq_poll.c to see how to write a poller.  There is a common API.  I'm not 100% certain that it is fully adequate, but if you find you need something that the core doesn't provide, let me know and we'll work it through.  Fundamentally, you have to implement these functions:

nni_posix_pollq_add() - adds a pollq node to the poller
nni_posix_pollq_remove() - removes a pollq node from the poller (usually just before it is destroyed)
nni_posix_pollq_arm() - arms events (which are poll(2) event items like POLLIN) on the node
nni_posix_pollq_disarm() - disarms events
nni_posix_pollq_get() - obtain a "pollq" for a given file descriptor.  The idea here was to support multiple pollq's for different CPUs, to get some cache coherence.  Right now I'm just returning a single global one.

When events that have been armed occur on the file descriptor associated with the node, you need to set the "revents" member and call the user supplied callback (cb).  Reading the posix_pollq_poll.c code should help you see what needs to be done.

The way I'd like to see the use of kqueue and the like enabled is via CMake tests,  for example look at how we detect the availability of clock_gettime() in the top level CMakeLists.txt.  This needs to enable a different NNG_USE_POSIX_POLLQ_KQUEUE variable, and the setting of the current in posix_config.h needs to  be made conditional, much like the clock definitions above it are done.

I had btw started doing this for epoll(), but set it aside for a bit.  I seem to recall that I did need to do something different there, and that I had some additional state requirements that had to be added to pollq_node, but I don't remember exactly what they were.

gdamore:
[...]

PR #204, opened 2018-01-15
kqueue: add kqueue-based pollq implementation

liamstask:
Initial cut at kqueue support for #35

This passes all tests on my machine, will see how it does on CI. Happy to rename/reorg as needed.

A couple thoughts:
* since some of the poller API (add, remove, enable, disable) may involve syscalls, it would nice to return errors from those routines and handle them accordingly, possibly in a future PR
* i wasn't totally clear on the expected handling of event flags - it looks like the existing poll implementation treats event flag manipulation as a cheap operation (it just manipulates userspace memory in the poll implementation), but since epoll and kqueue implementations might prefer to use the provided syscall interfaces to enable/disable events as distinct from adding/removing nodes, it could be nice to take that into consideration and minimize changes to enable/disable state.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -383,6 +383,7 @@ else ()
     nng_check_sym (backtrace_symbols_fd execinfo.h NNG_HAVE_BACKTRACE)
     nng_check_sym (alloca alloca.h NNG_HAVE_ALLOCA)
     nng_check_struct_member(msghdr msg_control sys/socket.h NNG_HAVE_MSG_CONTROL)
+    nng_check_sym (kqueue sys/event.h NNG_HAVE_KQUEUE)
 endif ()
 
 nng_check_sym (strlcat string.h NNG_HAVE_STRLCAT)
--- a/src/platform/posix/posix_pollq_kqueue.c
+++ b/src/platform/posix/posix_pollq_kqueue.c
@@ -1,0 +1,402 @@
+#ifdef NNG_HAVE_KQUEUE
+
+#include <errno.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <string.h> /* for strerror() */
+#include <sys/event.h>
+#include <unistd.h>
+
+#include "core/nng_impl.h"
+#include "platform/posix/posix_pollq.h"
+
+// TODO: can this be feature detected in cmake,
+//       rather than relying on platform?
+#if defined NNG_PLATFORM_NETBSD
+#define kevent_udata_t intptr_t
+#else
+#define kevent_udata_t void *
+#endif
+
+#define NNI_MAX_KQUEUE_EVENTS 64
+
+// user event id used to shutdown the polling thread
+#define NNI_KQ_EV_EXIT_ID 0xF
+
+// nni_posix_pollq is a work structure that manages state for the kqueue-based
+// pollq implementation
+struct nni_posix_pollq {
+	nni_mtx               mtx;
+	nni_cv                cv;
+	int                   kq;    // kqueue handle
+	bool                  close; // request for worker to exit
+	bool                  started;
+	nni_thr               thr;    // worker thread
+	nni_posix_pollq_node *wait;   // cancel waiting on this
+	nni_posix_pollq_node *active; // active node (in callback)
+};
+
+int
+nni_posix_pollq_add(nni_posix_pollq_node *node)
+{
+	nni_posix_pollq *pq;
+	struct kevent    kevents[2];
+
+	pq = nni_posix_pollq_get(node->fd);
+	if (pq == NULL) {
+		return (NNG_EINVAL);
+	}
+
+	// ensure node was not previously associated with a pollq
+	if (node->pq != NULL) {
+		return (NNG_ESTATE);
+	}
+
+	nni_mtx_lock(&pq->mtx);
+	if (pq->close) {
+		// This shouldn't happen!
+		nni_mtx_unlock(&pq->mtx);
+		return (NNG_ECLOSED);
+	}
+
+	node->pq     = pq;
+	node->events = 0;
+
+	EV_SET(&kevents[0], (uintptr_t) node->fd, EVFILT_READ,
+	    EV_ADD | EV_DISABLE, 0, 0, (kevent_udata_t) node);
+
+	EV_SET(&kevents[1], (uintptr_t) node->fd, EVFILT_WRITE,
+	    EV_ADD | EV_DISABLE, 0, 0, (kevent_udata_t) node);
+
+	if (kevent(pq->kq, kevents, 2, NULL, 0, NULL) != 0) {
+		nni_mtx_unlock(&pq->mtx);
+		return (nni_plat_errno(errno));
+	}
+
+	nni_mtx_unlock(&pq->mtx);
+	return (0);
+}
+
+// common functionality for nni_posix_pollq_remove() and nni_posix_pollq_fini()
+// called while pq's lock is held
+static void
+nni_posix_pollq_remove_helper(nni_posix_pollq *pq, nni_posix_pollq_node *node)
+{
+	int           rv;
+	struct kevent kevents[2];
+
+	node->events = 0;
+	node->pq     = NULL;
+
+	EV_SET(&kevents[0], (uintptr_t) node->fd, EVFILT_READ, EV_DELETE, 0, 0,
+	    (kevent_udata_t) node);
+
+	EV_SET(&kevents[1], (uintptr_t) node->fd, EVFILT_WRITE, EV_DELETE, 0,
+	    0, (kevent_udata_t) node);
+
+	rv = kevent(pq->kq, kevents, 2, NULL, 0, NULL);
+	// allow errnos that indicate the fd has already been removed
+	if (rv < 0 && errno != EBADF && errno != ENOENT) {
+		NNI_ASSERT(false);
+	}
+}
+
+// nni_posix_pollq_remove removes the node from the pollq, but
+// does not ensure that the pollq node is safe to destroy.  In particular,
+// this function can be called from a callback (the callback may be active).
+void
+nni_posix_pollq_remove(nni_posix_pollq_node *node)
+{
+	nni_posix_pollq *pq = node->pq;
+
+	if (pq == NULL) {
+		return;
+	}
+
+	nni_mtx_lock(&pq->mtx);
+	nni_posix_pollq_remove_helper(pq, node);
+
+	if (pq->close) {
+		nni_cv_wake(&pq->cv);
+	}
+	nni_mtx_unlock(&pq->mtx);
+}
+
+// nni_posix_pollq_init merely ensures that the node is ready for use.
+// It does not register the node with any pollq in particular.
+int
+nni_posix_pollq_init(nni_posix_pollq_node *node)
+{
+	NNI_ARG_UNUSED(node);
+	return (0);
+}
+
+// nni_posix_pollq_fini does everything that nni_posix_pollq_remove does,
+// but it also ensures that the callback is not active, so that the node
+// may be deallocated.  This function must not be called in a callback.
+void
+nni_posix_pollq_fini(nni_posix_pollq_node *node)
+{
+	nni_posix_pollq *pq = node->pq;
+	if (pq == NULL) {
+		return;
+	}
+
+	nni_mtx_lock(&pq->mtx);
+	while (pq->active == node) {
+		pq->wait = node;
+		nni_cv_wait(&pq->cv);
+	}
+
+	nni_posix_pollq_remove_helper(pq, node);
+
+	if (pq->close) {
+		nni_cv_wake(&pq->cv);
+	}
+	nni_mtx_unlock(&pq->mtx);
+}
+
+void
+nni_posix_pollq_arm(nni_posix_pollq_node *node, int events)
+{
+	nni_posix_pollq *pq = node->pq;
+	struct kevent    kevents[2];
+	int              nevents = 0;
+
+	NNI_ASSERT(pq != NULL);
+	if (events == 0) {
+		return;
+	}
+
+	nni_mtx_lock(&pq->mtx);
+
+	if (!(node->events & POLLIN) && (events & POLLIN)) {
+		EV_SET(&kevents[nevents++], (uintptr_t) node->fd, EVFILT_READ,
+		    EV_ENABLE | EV_DISPATCH, 0, 0, (kevent_udata_t) node);
+	}
+
+	if (!(node->events & POLLOUT) && (events & POLLOUT)) {
+		EV_SET(&kevents[nevents++], (uintptr_t) node->fd, EVFILT_WRITE,
+		    EV_ENABLE | EV_DISPATCH, 0, 0, (kevent_udata_t) node);
+	}
+
+	if (nevents > 0) {
+		int rv;
+		rv = kevent(pq->kq, kevents, nevents, NULL, 0, NULL);
+		NNI_ASSERT(rv >= 0);
+		node->events |= events;
+	}
+
+	nni_mtx_unlock(&pq->mtx);
+}
+
+void
+nni_posix_pollq_disarm(nni_posix_pollq_node *node, int events)
+{
+	struct kevent kevents[2];
+	int           nevents = 0;
+
+	nni_posix_pollq *pq = node->pq;
+	if (pq == NULL) {
+		return;
+	}
+
+	nni_mtx_lock(&pq->mtx);
+
+	if ((node->events & POLLIN) && (events & POLLIN)) {
+		EV_SET(&kevents[nevents++], (uintptr_t) node->fd, EVFILT_READ,
+		    EV_DISABLE, 0, 0, (kevent_udata_t) node);
+	}
+
+	if ((node->events & POLLOUT) && (events & POLLOUT)) {
+		EV_SET(&kevents[nevents++], (uintptr_t) node->fd, EVFILT_WRITE,
+		    EV_DISABLE, 0, 0, (kevent_udata_t) node);
+	}
+
+	if (nevents > 0) {
+		int rv = kevent(pq->kq, kevents, nevents, NULL, 0, NULL);
+		if (rv < 0 && errno != ENOENT && errno != EBADF) {
+			NNI_ASSERT(false);
+		}
+		node->events &= ~events;
+	}
+
+	nni_mtx_unlock(&pq->mtx);
+}
+
+static void
+nni_posix_poll_thr(void *arg)
+{
+	nni_posix_pollq *pq = arg;
+	struct kevent    kevents[NNI_MAX_KQUEUE_EVENTS];
+
+	nni_mtx_lock(&pq->mtx);
+
+	while (!pq->close) {
+		int i;
+		int nevents;
+
+		// block indefinitely, timers are handled separately
+		nni_mtx_unlock(&pq->mtx);
+		nevents = kevent(
+		    pq->kq, NULL, 0, kevents, NNI_MAX_KQUEUE_EVENTS, NULL);
+		nni_mtx_lock(&pq->mtx);
+
+		if (nevents < 0) {
+			continue;
+		}
+
+		// dispatch events
+		for (i = 0; i < nevents; ++i) {
+			struct kevent         ev_disable;
+			const struct kevent * ev;
+			nni_posix_pollq_node *node;
+
+			ev = &kevents[i];
+			if (ev->filter == EVFILT_USER &&
+			    ev->ident == NNI_KQ_EV_EXIT_ID) {
+				// we've woken up to exit the polling thread
+				break;
+			}
+
+			node = (nni_posix_pollq_node *) ev->udata;
+			if (node->pq == NULL) {
+				// node was removed while we were blocking
+				continue;
+			}
+			node->revents = 0;
+
+			if (ev->flags & (EV_ERROR | EV_EOF)) {
+				node->revents |= POLLHUP;
+			}
+			if (ev->filter == EVFILT_WRITE) {
+				node->revents |= POLLOUT;
+			} else if (ev->filter == EVFILT_READ) {
+				node->revents |= POLLIN;
+			} else {
+				NNI_ASSERT(false); // unhandled filter
+				break;
+			}
+
+			// explicitly disable this event. we'd ideally rely on
+			// the behavior of EV_DISPATCH to do this,
+			// but that only happens once we've acknowledged the
+			// event by reading/or writing the fd. because there
+			// can currently be some latency between the time we
+			// receive this event and the time we read/write in
+			// response, disable the event in the meantime to avoid
+			// needless wakeups.
+			// revisit if we're able to reduce/remove this latency.
+			EV_SET(&ev_disable, (uintptr_t) node->fd, ev->filter,
+			    EV_DISABLE, 0, 0, NULL);
+			// this will only fail if the fd is already
+			// closed/invalid which we don't mind anyway,
+			// so ignore return value.
+			kevent(pq->kq, &ev_disable, 1, NULL, 0, NULL);
+
+			// mark events as cleared
+			node->events &= ~node->revents;
+
+			// Save the active node; we can notice this way
+			// when it is busy, and avoid freeing it until
+			// we are sure that it is not in use.
+			pq->active = node;
+
+			// Execute the callback with lock released
+			nni_mtx_unlock(&pq->mtx);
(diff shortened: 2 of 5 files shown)


src/platform/posix/posix_epdesc.c
```diff
@@ -103,7 +103,7 @@ nni_posix_epdesc_doconnect(nni_posix_epdesc *ed)
 		case 0:
 			// Success!
 			nni_posix_epdesc_finish(aio, 0, ed->node.fd);
-			ed->node.fd = -1;
+			// ed->node.fd = -1;
```
liamstask: i'm not sure this is an appropriate change, but without it I was seeing the httpclient test fail during cleanup - because the fd was set to `-1` but not closed, epdesc cleanup code doesn't close it (it's already `-1`) but the poller would deliver a callback after the epdesc had been cleaned up, causing an assert when accessing its mutex.

Please let me know if there's a better way to address this.

Relatedly, in `nni_posix_epdesc_finish()` I don't immediately see where the `nni_posix_pipedesc` that can be allocated there gets free'd - may be that i'm just not familiar with the ownership system there, but could possibly be leaking?

CMakeLists.txt
```diff
@@ -324,6 +324,7 @@ macro (nng_check_sym SYM HDR DEF)
     check_symbol_exists (${SYM} ${HDR} ${DEF})
     if (${DEF})
         add_definitions (-D${DEF}=1)
+        set (${DEF} ON)
```
gdamore: This should not be needed.  CMake should already have set that with check_symbol_exists.

liamstask: it's needed in order to conditionalize other cmake inclusions, like the one in src/CMakeLists.txt - let me know if you'd prefer another approach.

gdamore: Let me look at this in more detail -- I still think this is redundant unless I've misunderstood what check_symbol_exists does.

gdamore:
I need to review this, and probably won't have time to complete the review for a couple of days.  At a high level, I suspect we will want to make some minor file reorganizations at least.  (For example, I consider "bsd" as subset of posix, as it uses all the posix stuff for ~everything else, so I don't want to create a new subdirectory for it.  Also you're already doing the right thing by guarding the code in the kqueue poller, so we might just add both files unconditionally, and assume that they got the guards set properly.)

More as I have time to read.

liamstask:
Sounds good. The reason i originally added files conditionally within cmake was the following (benign) warning:

```
[135/270] Linking C static library libnng_static.a
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libnng_static.a(posix_pollq_poll.c.o) has no symbols
```

Happy to use that approach if you prefer.

gdamore:
So the way I've solved that warning in the past is creating a bogus symbol, e.g. "int posix_pollq_kqueue_notused = 0;"

Now having said that, I am leaning more and more towards using CMake here like you've done, but I think I may want to move the "platform" logic into a new cmake tree.   Stay tuned for that.

src/platform/posix/posix_epdesc.c
gdamore: I think I may want to refactor some of this somewhat.  Testing with helgrind is actually highlighting that the way we are tweaking the node.fd is actually a potential race.  

So with respect to the pipedesc, if the pipedesc creation succeeds, then rv will be zero, and we will pass that out to the caller in the nni_aio_finish_pipe() logic.  The aio callback then has the responsibility to "own" the pipe, and generally this causes the pipe to get associated with the socket.  That logic is handled in the upper core code, endpt.c and pipe.c.  I don't believe there are any leaks here, and valgrind at least agrees with me in that regard.

gdamore:
So the other thing is that I'm a bit uncomfortable with "ignoring" system calls, unless we are reasonably certain that they could not ever fail.  Unfortunately, there are numerous reasons these calls can fail (ENOMEM, EACCES, EINTR) which the calling program may not have control over.

(I'm ok with ignoring failures from system calls that can only fail if we provide bogus input, like EFAULT due to passing a bad pointer -- since in theory we can reasonably be assured that our own code is not buggy in this regard.)  Although in that case an assertion may be called for.

I need to think about the arm/disarm API, because those lack a way to pass event updates back.  It would be better if we didn't need to pass such updates (having an API that is immune to failure is better!) but I'm not sure we can do that.

Stay tuned.

liamstask:
Agreed, same concern noted in the pr description above 😄

Also, I wonder if it would be worth reconsidering the current approach of disarming and re-arming nodes after each event - could save a bit of work by keeping them armed until they're explicitly requested to be disarmed, assuming that would be relatively infrequent (currently looks like it's just done during shutdown of an endpoint.

src/platform/posix/posix_pollq_kqueue.c
```diff
@@ -0,0 +1,309 @@
...
+}
+
+static int
+nni_bsd_kqueue_init_helper(nni_posix_pollq *pq)
+{
+	int           rv;
+	struct kevent event;
+
+	if ((rv = nni_plat_pipe_open(&pq->wakewfd, &pq->wakerfd)) != 0) {
+		return rv;
```
gdamore: Stylistic issue here: I prefer return values always be surrounded by parens.

src/platform/posix/posix_pollq_kqueue.c
```diff
@@ -0,0 +1,309 @@
...
+		EV_SET(&kevents[nevents++], node->fd, EVFILT_WRITE, EV_DELETE,
+		    0, 0, (kevent_udata_t) node);
+	}
+
+	if (nevents > 0) {
+		int rv;
+		rv = kevent(pq->kq, kevents, nevents, NULL, 0, NULL);
+		if (rv < 0) {
+			// report error?
+			fprintf(stderr, "disarm err: %s (%d)\n",
```
gdamore: So it turns out that we currently only "disarm" events during "close".   Apparently close() of a fd implicitly unregisters the associated kevents:

From kevent(2) on macOS:

>  Calling close() on a file descriptor will remove any kevents that reference the descriptor.

I'm looking at the other side next.

gdamore: Btw, what I mean here is that we should just ignore the return code, and inject a comment to this effect.

src/platform/posix/posix_pollq_kqueue.c
```diff
@@ -0,0 +1,309 @@
...
+
+	if (!(node->events & POLLOUT) && (events & POLLOUT)) {
+		EV_SET(&kevents[nevents++], node->fd, EVFILT_WRITE, EV_ADD, 0,
+		    0, (kevent_udata_t) node);
+	}
+
+	if (nevents > 0) {
+		int rv;
+		rv = kevent(pq->kq, kevents, nevents, NULL, 0, NULL);
+		if (rv < 0) {
```
gdamore: Looking at this, I'm inclined to think what we ought to do is mark the event as if POLLERR was triggered.  It would be even handier if we could stash the errno somewhere, or some indication that it was *kevent* that triggered the error.  Stay tuned.

src/platform/posix/posix_pollq_kqueue.c
```diff
@@ -0,0 +1,309 @@
...
+		int i;
+		int nevents;
+
+		// block indefinitely, timers are handled separately
+		nni_mtx_unlock(&pq->mtx);
+		nevents = kevent(pq->kq, NULL, 0, kevents, maxevents, NULL);
+		nni_mtx_lock(&pq->mtx);
+
+		if (nevents < 0) {
+			// report error?
```
gdamore: This is an unfortunate place to have this kind of error... I'm not sure how best to deal with it....  can this really fail?  My read of the man page makes me thing any error other than EINTR here is impossible.

gdamore: The more I think about this the more I think we need to have a new error flag or state on the node, and let the poller set this.  If we have an error here, we would trigger that error on *every* item in the pollq.

This may help other poller implementations that have a need to report some other kind of weird failure scenario.

liamstask:
I started experimenting with a slightly different approach, using the following mapping:
* `EV_ADD` -> `nni_posix_pollq_add()`
* `EV_DELETE` -> `nni_posix_pollq_remove()`
* `EV_ENABLE` -> `nni_posix_pollq_arm()`
* `EV_DISABLE` -> `nni_posix_pollq_disarm()`

which arguably makes more sense. I think this would make a better case for ignoring errors on arm/disarm since those operations should essentially consist of setting a flag internally that should already be allocated.

one issue i quickly ran into - `node->fd` is 0 when `nni_posix_pollq_add()` is called, at least for tcp, such that the fd can't actually be added.

can special case this, of course, by checking for whether we've seen a valid fd by the time we're calling arm/disarm, but not ideal. if that's intended, do you have any suggestions for handling that?

gdamore:
I seem to recall diving into the kernel code for kevent and finding cases
where even enable and disable can fail due to ENOMEM. I will review it
more.

The tcp thing is due to either connect or accept (dont recall which) but
will look more closely shortly.

gdamore:
So I've gone through the FreeBSD code, and it appears that only EV_ADD can generate ENOMEM.  The other *documented* errors seem to be impossible for correctly written code to cause ... (EBADF, EFAULT, EACCES, EINVAL, ENOENT, ESRCH) with the possible exception of EINTR.  I need to go through your code and examine that more closely to see if it can happen.

The endpoint descriptors call posix_pollq_add to create the pollq node before the file descriptor is actually opened!

Probably this needs to be done as part of the posix_epdesc_listen or posix_epdesc_connect calls.
The alternative would be to open the underlying socket descriptor (unbound) during the epdesc initialization.

We don't have this problem with the pipedesc.

gdamore:
I'm going to try to change the code to ensure that we only call the _add() function when we have a corresponding file descriptor.  Stay tuned.

gdamore:
Please have a look at the changes in the "epdesc" branch.  This separates the pollq_add from initialization, so that pollq_add is only ever called when the node has a valid file descriptor.  It also does some work to clean up the connect side, so that we do pollq_remove() when the connect logic finishes.  (It will subsequently be re-added as part of a pipedesc rather than an endpoint desc.)

I think these changes, combined with your proposal to use EV_ENABLE and EV_DISABLE in the arm and disarm routines, should lead us to a nice solution.

liamstask:
That sounds 👍 to me - looks like the `epdesc` is not pushed yet so cannot see it, but once you do, I can try to rebase a version of this PR using EV_ENABLE/EV_DISABLE on that.

liamstask:
just rebased on master & added a commit that implements the approach discussed above. (happy to squash/rebase as needed prior to merge.)

this passes all tests except ipc - i am still debugging an issue there, but thought i would push to continue the discussion and see if you had any pointers :)

issue is that i am seeing a sequence of calls for a given node:
1. nni_posix_pollq_add()
1. nni_posix_pollq_arm()
1. [event is delivered]
1. nni_posix_pollq_remove()
1. nni_posix_pollq_arm() - this fails as the node has already been removed

all other tests are passing, so will continue to gather details on this and run it down.

gdamore:
I will have a look at the posix pollq_arm vs. not added flow... my guess there is that I botched something in the connection setup flow.

src/platform/posix/posix_epdesc.c
```diff
@@ -104,7 +104,7 @@ nni_posix_epdesc_doconnect(nni_posix_epdesc *ed)
 			// Success!
 			nni_posix_pollq_remove(&ed->node);
 			nni_posix_epdesc_finish(aio, 0, ed->node.fd);
-			ed->node.fd = -1;
+			// ed->node.fd = -1;
```
gdamore: Lets just remove it altogether, please.

src/platform/posix/posix_epdesc.c
```diff
@@ -131,9 +131,9 @@ nni_posix_epdesc_doaccept(nni_posix_epdesc *ed)
 	int      newfd;
 
 	while ((aio = nni_list_first(&ed->acceptq)) != NULL) {
-// We could argue that knowing the remote peer address would
-// be nice.  But frankly if someone wants it, they can just
-// do getpeername().
+		// We could argue that knowing the remote peer address would
+		// be nice.  But frankly if someone wants it, they can just
+		// do getpeername().
```
gdamore: This goofy indentation (my old one I mean) may have been done automatically by clang-format.  In which case it may wind up getting reverted later.  I don't always agree 100% with the decisions it makes, but I find using it consistently means I don't have to get into debates about formatting style.

Still it isn't as aggressive about undoing changes as some other formatting tools, so its possible that this change won't get trashed.   This is just a heads up in case it does.

src/platform/posix/posix_pollq_kqueue.c
```diff
@@ -0,0 +1,392 @@
...
+#include <errno.h>
+#include <stdio.h>
+#include <string.h> /* for strerror() */
+#include <sys/event.h>
+#include <unistd.h>
+
+#include "core/nng_impl.h"
+#include "platform/posix/posix_pollq.h"
+
+#if defined NNG_PLATFORM_NETBSD
```
gdamore: It would be nice to have this detected by a check, rather than using the platform.  I'm not sure how you can test the type of a field member under cmake though.

src/platform/posix/posix_pollq_kqueue.c
```diff
@@ -0,0 +1,392 @@
...
+}
+
+static void
+nni_posix_pollq_destroy(nni_posix_pollq *pq)
+{
+	if (pq->started) {
+		nni_mtx_lock(&pq->mtx);
+		pq->close   = 1;
+		pq->started = 0;
+		nni_plat_pipe_raise(pq->wakewfd);
```
gdamore: I think we can probably dispense with the wakewfd and wakerfd and the plat pipe logic.  That logic exists because I don't have any other way to alert a pollq using poll to a change in the polled set of descriptors.  I think kqueue doesn't suffer from this problem at all, since it uses explicit system calls for this.

gdamore: You do need to do something to cause the poll thread to notice and exit though, so maybe this pipe is still useful?  Alternatively what happens if you just close the kqueue fd?  Does the thread wake up?  Is there some other convenient way to force an event via the kqueue system call?

src/platform/posix/posix_pollq_kqueue.c
```diff
@@ -0,0 +1,392 @@
...
+		return (rv);
+	}
+	if ((rv = nni_thr_init(&pq->thr, nni_posix_poll_thr, pq)) != 0) {
+		return (rv);
+	}
+
+	// always add wake fd so we can be signaled
+	EV_SET(
+	    &event, (uintptr_t) pq->wakerfd, EVFILT_READ, EV_ADD, 0, 0, NULL);
+	return kevent(pq->kq, &event, 1, NULL, 0, NULL);
```
gdamore: style nit: please use return (value) // note the parens

src/platform/posix/posix_pollq_kqueue.c
```diff
@@ -0,0 +1,392 @@
...
+static int
+nni_posix_pollq_init_helper(nni_posix_pollq *pq)
+{
+	int           rv;
+	struct kevent event;
+
+	if ((rv = nni_plat_pipe_open(&pq->wakewfd, &pq->wakerfd)) != 0) {
+		return (rv);
+	}
+	if ((rv = nni_thr_init(&pq->thr, nni_posix_poll_thr, pq)) != 0) {
```
gdamore: I'm wondering if we should not be creating *multiple* of these threads, to increase scalability.  I know with the posix poller I only create "one" at present, but I intended to scale that across multiple pollq instances.  With kqueue it might make more sense to have just a single global pollq and run multiple thread instances.  (Some assessment about lock contention may be desirable though -- having separate pollq and kqueue instances would eliminate any need for locking and increase locality I suppose.)

src/platform/posix/posix_epdesc.c
gdamore: Actually, I believe this code needs to stay the way it was.  Now that we are *removing* the pollq, the value of the fd is -1, and this ensures that we do not attempt to reuse the same descriptor for connect, but instead allocate a new one.   I'm wondering if this is the source of your grief about remove followed by arm...

src/platform/posix/posix_epdesc.c
gdamore: Looking at this more closely, we should remove this comment ... its no longer relevant due to other changes in the code.

gdamore:
So I figured out the problem.  I'll be pushing a fix, but the issue is that we sometimes close the descriptor from the callback.  We therefore cannot automatically rearm, because it may have been closed (and removed).  I'm also adding a check against events being 0, which is kind of redundant, but it can save some expensive operations (getting a lock) when we don't have anything new to arm.

gdamore:
Have a look at https://github.com/nanomsg/nng/pull/242 -- I believe that this will address the problem you're seeing.  I intend to merge that if CI comes back clean.

gdamore:
So, that worked, but now you'll have to rebase, sorry.  Please also review my comments above -- you don't want to remove the reset of the node.fd for example.

The override of DEFS in CMake files should be removed -- I'm convinced it is redundant.

I'm ok with deferring any changes related to scaling to multiple threads, or the change to the wake pipe, until after this is merged.  (I really hope we can ditch the wake pipes though -- I think they add complexity that its likely we do not need.)

We're really close I think.

src/platform/posix/posix_pollq_kqueue.c
liamstask: hm yea - i just grabbed this from nanomsg & have not tested on net bsd specifically. i'll add a TODO comment and leave it as is for now.

src/platform/posix/posix_pollq_kqueue.c
liamstask: closing the kqueue is enough to wake it up and exit - will remove wake fds for now, can always bring them back if needed.

liamstask:
alright - rebased on your latest changes and everything's looking 👌all tests are passing here on 10's of runs.

removed the wake fd handling and the cmake override - confirmed that cmake's `check_symbol_exists` creates the same variable already.

i also added a minor cleanup of ipc init that i stumbled across while debugging - not the source of the problem but seemed worth including.

gdamore:
This looks really quite good.  I'm going to do some quick performance runs on my mac, but I expect to merge this shortly.  Thanks for your hard work on this!

gdamore:
Well, my performance runs gave disappointing results.  On the one hand I didn't really expect a *boost* because the test I have handy is single threaded, but I see a substantial performance regression in the throughput test (and a much smaller regression in the latency test.)

I think the problem is that the cost of the system calls is hurting us badly.  I expect the code to *scale* better, but I want to eyeball this again to see if there is any place we can get some wins... one though that occurred to me is that if we use ONESHOT, maybe we can dispense with the calls to disable the node.  Since we're doing this ourselves with a lock, we could easily keep track of the enabled vs. disabled state and simply not make the calls that we don't need.

The other thing to look at more closely is our use of the lock itself -- is it still needed?  I doubt it has much affect (it shouldn't be contended), but every little bit counts.

If you get time to looks at this more closely, maybe you can give your thoughts.

One other idea might be to make the use of kqueue optional, so that users can fine tune for either single threaded/socket performance, or greater scalability, since that does seem to be the major tradeoff here.

liamstask:
hm! can you share any details on your perf test? if i can reproduce it here, i can try to profile a bit and see if there's any reasonably low hanging fruit.

at least according to comments in the header, `EV_DISPATCH` claims to disable the event after reporting (and subsequent calls to pollq_disable() should avoid the system call), but perhaps there are some other notification/filter options that might be advantageous? can explore a bit.

liamstask:
one problem i've seen so far is that there is sometimes nontrivial latency between the initial event notification (ie, event returned from kqueue) and when the actual io takes place (ie, `readv()`/`writev()` is called).

this happens because the `EV_DISPATCH` flag, which we rely on to automatically disable the event, only disables the event after it's been ack'd by the user (ie, the fd has been read/written).

explicitly making a call to disable (via `kevent()`) prior to invoking the cb addresses this issue, at the cost of an additional syscall, but it may also suggest there's some other undesirable behavior (contention somewhere?) causing this delay in handling the reported events.

from running several of the tests, the worst instances i've seen are in the multistress test where 250 event notifications occur before the related io occurs.

that said, not sure exactly what perf tests you are running, so difficult to say whether this could account for the issues you've seen, but it certainly results in the system doing a lot of extra/wasted work.

let me know how you'd like to proceed - i can include the simple change that explicitly disables before callbacks and see if this improves the numbers that you see, or continue investigating.

liamstask:
I see similar behavior (ie, multiple notifications) with `EV_ONESHOT`. my reading of the man page on that is that it will behave similarly to `EV_DISPATCH` but instead of disabling the event after it's been consumed, it will delete it, such that it would need to be added again, rather than just re-enabling. maybe a closer look at the source is needed.

thanks for the pointer - i'll check out the perf tests and compare them to the `poll()` based implementation.

liamstask:
after some brief testing, it looks like this issue may indeed account for the drop in throughput. I did 2 runs each for the throughput and latency tests in the following configurations:
* without the explicit disable after event notification (ie, the code as it existed at f9a81bd prior to the most recent 2 commits)
* with the explicit disable (ie latest commit at 6854d8d)
* with kqueue support disabled altogether, relying on poll()

obviously not statistically significant, but on first glance It looks like throughput of the explicit disable variant  is more or less the same as poll(), and latency is slightly improved.

let me know if you're able to reproduce.

```
***************** kqueue() without explicit disable

total time: 0.052 [s]
message size: 1024 [B]
message count: 500
throughput: 9615 [msg/s]
throughput: 75.120 [Mb/s]

total time: 0.051 [s]
message size: 1024 [B]
message count: 500
throughput: 9804 [msg/s]
throughput: 76.593 [Mb/s]


total time: 0.115 [s]
message size: 1024 [B]
round trip count: 500
average latency: 115.000 [us]

total time: 0.115 [s]
message size: 1024 [B]
round trip count: 500
average latency: 115.000 [us]

***************** kqueue() with explicit disable

total time: 0.046 [s]
message size: 1024 [B]
message count: 500
throughput: 10870 [msg/s]
throughput: 84.918 [Mb/s]

total time: 0.046 [s]
message size: 1024 [B]
message count: 500
throughput: 10870 [msg/s]
throughput: 84.918 [Mb/s]


total time: 0.106 [s]
message size: 1024 [B]
round trip count: 500
average latency: 106.000 [us]

total time: 0.105 [s]
message size: 1024 [B]
round trip count: 500
average latency: 105.000 [us]

***************** poll()

total time: 0.045 [s]
message size: 1024 [B]
message count: 500
throughput: 11111 [msg/s]
throughput: 86.806 [Mb/s]

total time: 0.046 [s]
message size: 1024 [B]
message count: 500
throughput: 10870 [msg/s]
throughput: 84.918 [Mb/s]


total time: 0.107 [s]
message size: 1024 [B]
round trip count: 500
average latency: 107.000 [us]

total time: 0.117 [s]
message size: 1024 [B]
round trip count: 500
average latency: 117.000 [us]
```

gdamore:
Actually it looks like the latency with explicit disable dropped slightly?  Anyway, as this is single threaded performance, as long as we are getting close to poll(), then I'm pretty happy.  The purpose of kqueue() is not to make single threaded performance go faster (it doesn't), but to scale much better by elimination of expensive walks through long event lists that are required with poll().

I'll test this out myself and get back to you.  If you rebase to HEAD, you can also eliminate your changes to the perf tool -- I have the same change in HEAD.

liamstask:
rebased on current master, squashed the explicit disable commit into the main kqueue commit & also added functionality to explicitly wake up poller (via `EVFILT_USER`) rather than the brute force approach of just `close()`ing the kqueue.

passing tests and perf appears to be similar to poll() - i believe it's ready to merge if perf looks ok on your end.

gdamore:
I'm happy with the test results.  Thanks for doing this.

gdamore:
One thing I've belatedly noticed, is that you don't have a copyright notice on your work.  I'd like to correct that, using the same license header, but your own copyright.

gdamore:
Its important that this work be licensed under MIT the same as the rest of the code.

I'd propose adding the following to the kqueue C file:

```
//
// Copyright 2018 Liam Staskawicz <liam@stask.net>
//
// This software is supplied under the terms of the MIT License, a
// copy of which should be located in the distribution where this
// file was obtained (LICENSE.txt).  A copy of the license may also be
// found online at https://opensource.org/licenses/MIT.
//
```

Additionally if you copy-pasted any code from the poll.c file, it should have Staysail and Capitar copyrights as well.

PR #259, opened 2018-03-03
update ZeroTier transport man page to account for URI change

### LINKED ISSUE #258: ZeroTier URI Format documentation change
Need to update documentation to account for URI change.

Old zt://nodeid.nwid:port

New zt://nwid.nodeid:port

janjaapbos:
fixes #258 ZeroTier URI Format documentation change

### DIFF
--- a/docs/man/nng_zerotier.adoc
+++ b/docs/man/nng_zerotier.adoc
@@ -57,11 +57,11 @@ cannot be initialized for any reason.
 
 === URI Format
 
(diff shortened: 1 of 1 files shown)


gdamore:
Actually, the documentation is correct.  The nodeid is parsed before the network id.  I just checked the code.  We changed this a while back so that the parsing would work more like network names -- <host>.<domain>   -- so we have <node>.<network>.

gdamore:
Closing this as the docs are correct as is.

PR #264, opened 2018-03-05
epoll: add epoll-based pollq implementation

### LINKED ISSUE #33: Add epoll() support on Linux
We need an epoll() based poller to increase scalability.

liamstask:
fixes #33

This is ultimately pretty similar to the kqueue PR, main differences are:
* the epoll oneshot flag behavior allows us to avoid an explicit call to disable events as in kqueue, as they are disabled as soon as they are reported via `epoll_wait()`
* using the pipe fds to wake the poll thread in destroy appears to be necessary, as just closing the `epfd` doesn't work

Other than that, I think most of the integration effort for the kqueue PR made this one substantially more straightforward :)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -108,6 +108,10 @@ if (NNG_HAVE_KQUEUE)
     set (NNG_SOURCES ${NNG_SOURCES}
         platform/posix/posix_pollq_kqueue.c
     )
+elseif (NNG_HAVE_EPOLL)
+    set (NNG_SOURCES ${NNG_SOURCES}
+        platform/posix/posix_pollq_epoll.c
+    )
 else()
     set (NNG_SOURCES ${NNG_SOURCES}
         platform/posix/posix_pollq_poll.c
--- a/src/platform/posix/posix_pollq_epoll.c
+++ b/src/platform/posix/posix_pollq_epoll.c
@@ -1,0 +1,379 @@
+//
+// Copyright 2018 Staysail Systems, Inc. <info@staysail.tech>
+// Copyright 2018 Capitar IT Group BV <info@capitar.com>
+// Copyright 2018 Liam Staskawicz <liam@stask.net>
+//
+// This software is supplied under the terms of the MIT License, a
+// copy of which should be located in the distribution where this
+// file was obtained (LICENSE.txt).  A copy of the license may also be
+// found online at https://opensource.org/licenses/MIT.
+//
+
+#ifdef NNG_HAVE_EPOLL
+
+#include <errno.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <string.h> /* for strerror() */
+#include <sys/epoll.h>
+#include <sys/eventfd.h>
+#include <unistd.h>
+
+#include "core/nng_impl.h"
+#include "platform/posix/posix_pollq.h"
+
+#define NNI_MAX_EPOLL_EVENTS 64
+
+// flags we always want enabled as long as at least one event is active
+#define NNI_EPOLL_FLAGS (EPOLLONESHOT | EPOLLERR | EPOLLHUP)
+
+// nni_posix_pollq is a work structure that manages state for the epoll-based
+// pollq implementation
+struct nni_posix_pollq {
+	nni_mtx               mtx;
+	nni_cv                cv;
+	int                   epfd;  // epoll handle
+	int                   evfd;  // event fd
+	bool                  close; // request for worker to exit
+	bool                  started;
+	nni_thr               thr;    // worker thread
+	nni_posix_pollq_node *wait;   // cancel waiting on this
+	nni_posix_pollq_node *active; // active node (in callback)
+};
+
+int
+nni_posix_pollq_add(nni_posix_pollq_node *node)
+{
+	int                rv;
+	nni_posix_pollq *  pq;
+	struct epoll_event ev;
+
+	pq = nni_posix_pollq_get(node->fd);
+	if (pq == NULL) {
+		return (NNG_EINVAL);
+	}
+
+	// ensure node was not previously associated with a pollq
+	if (node->pq != NULL) {
+		return (NNG_ESTATE);
+	}
+
+	nni_mtx_lock(&pq->mtx);
+	if (pq->close) {
+		// This shouldn't happen!
+		nni_mtx_unlock(&pq->mtx);
+		return (NNG_ECLOSED);
+	}
+
+	node->pq     = pq;
+	node->events = 0;
+
+	// notifications disabled to begin with
+	ev.events   = 0;
+	ev.data.ptr = node;
+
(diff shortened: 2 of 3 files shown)


gdamore:
Wow, I had planned to do this tomorrow actually.   I will review and test your changes at least then! :-)

src/CMakeLists.txt
```diff
@@ -104,7 +104,11 @@ if (NNG_PLATFORM_POSIX)
     )
 endif()
 
-if (NNG_HAVE_KQUEUE)
+if (NNG_HAVE_EPOLL)
```
gdamore: I actually think kqueue should have precedence.  Some systems are emulating epoll, and I won't be surprised if BSDs do that too to facilitate portability, but kqueue is definitely better.  epoll() itself is rather defective under the hood, and we would prefer not to use it unless no other alternative is viable.

src/platform/posix/posix_pollq_epoll.c
```diff
@@ -0,0 +1,372 @@
...
+
+	pq->wakewfd = -1;
+	pq->wakerfd = -1;
+	pq->close   = false;
+
+	nni_mtx_init(&pq->mtx);
+	nni_cv_init(&pq->cv, &pq->mtx);
+
+	if (((rv = nni_thr_init(&pq->thr, nni_posix_poll_thr, pq)) != 0) ||
+	    ((rv = nni_posix_pollq_add_wake_pipe(pq)) != 0)) {
```
gdamore: Thinking about this, rather than using the clunky wakewfd, wakerfd, why not just go ahead and use eventfd() directly.  We know its here, because we're on Linux.  (We shouldn't use epoll() anywhere else.)

src/platform/posix/posix_pollq_epoll.c
```diff
@@ -0,0 +1,372 @@
...
+	nni_posix_pollq *pq = node->pq;
+	if (pq == NULL) {
+		return;
+	}
+
+	nni_mtx_lock(&pq->mtx);
+
+	node->events &= ~events;
+	if (node->events == 0) {
+		ev.events = 0;
```
gdamore: I wonder if we actually need to make the system call in this case.  We have the ONESHOT flag, so we shouldn't keep getting called here....

src/platform/posix/posix_pollq_epoll.c
liamstask: 👍 implemented in 7d8e166

src/CMakeLists.txt
liamstask: done in 7d8e166

src/platform/posix/posix_pollq_epoll.c
liamstask: the only thing i can think of is the case in which an event has been armed, and disarm is called before any events have been delivered. this could result in the next event being delivered despite disarm having been called, since we wouldn't have actively disabled it.

if callers can deal with that, then agreed we should be able to avoid the call to `epoll_ctl()` here.

existing tests pass with your proposed change - let me know which you prefer.

gdamore:
The http server test faulted on Linux.  I'm not quite ready to sweep this under the table, since we've never seen it anywhere else, and the code you changed is quite possibly relevant.  I'll look and see if I can turn up anything, but it might be a day or two.

liamstask:
makes sense. i pushed #270 to help capture more info for issues like this in the future.

interestingly, i cannot reproduce any failures when running the test binary directly, ie `./tests/httpserver` from the build dir.

however, i periodically see failures in the httpserver test when running via ctest (which is invoked via `ninja test`, which is used in ci) returning `NNG_EADDRINUSE` for nng_http_server_start()`, ie:

```
root@3321c08eb0fc:/nng/linux-build# env CTEST_OUTPUT_ON_FAILURE=1 ctest -R httpserver
Test project /nng/linux-build
    Start 8: httpserver
1/1 Test #8: httpserver .......................***Failed    0.54 sec
```

If I wait long enough between test runs, I don't see this failure, as expected. Maybe SO_REUSEADDR/SO_REUSEPORT could be of assistance?

However, this doesn't appear to correspond to the case that occurred on that travis build, since the print out in that case was `***Exception: Other` rather than the `***Failed` I've seen above.

I haven't been able to turn up anything conclusive on `Exception: Other`/`OTHER_FAULT` but the cases in which I've seen it mentioned are all related to dynamic linking issues, an exe not being able to find a .so or something along those lines.

Not sure how that could be the case here unless there's something strange going on w Travis :/

So if some form of #270 gets merged, perhaps I'll just rebase on top of that and rebuild several more times to see if it reproduces.

gdamore:
I've added that change... we'll see if it helps with diagnosis.

liamstask:
hm, seems to have passed this time :/

From a scan of the log output on [this passing run](https://travis-ci.org/nanomsg/nng/jobs/350655297) compared to [this failing run](https://travis-ci.org/nanomsg/nng/jobs/349661097) everything looks pretty much the same, two differences i noted:
* under the `Build system information` heading, the `travis-build` id is different (not sure what that corresponds to)
* in the passing build it prints "Network availability confirmed." which is not printed in the failing build

probably not enough to blame it on travis, yet. maybe once you get a chance to test it locally, we can go from there.

liamstask:
hm, 2 more passes ([#882](https://travis-ci.org/nanomsg/nng/builds/351159587) and [#883](https://travis-ci.org/nanomsg/nng/builds/351598453) on travis) after rebasing on master a couple times.

given that the original error reported does not correspond to a test assertion failure or a segfault, but some "OTHER_FAULT" exception, I'm starting to suspect something environmental on that run.

liamstask:
[#908](https://travis-ci.org/nanomsg/nng/jobs/352650911) is an interesting failure - when i run `./tests/aio` locally, I don't see any failures, but I also notice that it does not add any nodes to the poller at all, suggesting that it might not be caused by the epoll changes. any other ideas about how they might be related?

liamstask:
@gdamore ping - thoughts on this PR? I haven't changed any code since responding to your initial feedback, have just been rebasing periodically since to trigger more runs.

gdamore:
i will take some time for it tomorrow. stay tuned.

gdamore:
Looking over this, and doing some testing, things look reasonably good.

I saw some problems, but I think I've verified they are in my Ubuntu setup (and combined with the way we test for invalid URLs).

The failure in 908 is due to a race condition and bogus scheduling assumption -- it's bad test code, and I believe I've fixed that now.  (The test app was making the bad assumption; what was happening was that the receive callback was executed before the send callback.  Normally you wouldn't expect this, but there's nothing to say that a scheduler can't do that if the send/recv occurs fast enough.)

I think it's likely that the server fault in the earlier bug was indeed collisions in the address use, so I'm discounting that now.

I'll do some performance tests, to make sure this doesn't regress, then assuming it's good, I'll merge it.

PR #292, opened 2018-03-16
Proposed small fixes to align .h headers and implementations

nonnenmacher:
To allow making proper C binding to Python, the .h declarations should match what is visible in the shared library when `dlopening` it from Python.

Some  entries (e.g `nng_aio_abort`) are declared as extern, but there is no implementation or some seems small typos.

As in python it is harder to parse includes and properly handle #define (without resorting to a full parser), it is easier to have 'empty/null' implementation of functions *parsed* as `extern` definitions from .h

`nng_snapshot_xxx` and `nng_stats_xxx` functions are in this case. 
Is this correct to let them goes into the librairies as they are harmless ?

Last, perhaps more problematic one, is on the same line of non been able to handle `#define/#ifdef..` as there is a declaration for an `extern int nng_wss_register(void)`, the python binding expect to find one in the library and stop if if doesn't find one.

If this is correct for you, the proposed modifications allow to make a `cffi` binding for the new `nng` lib.


nonnenmacher:
I have commented the declarations in nng.h and everything is fine on the Python cffi side.

gdamore:
OOPs...
I'm going to back that out... i meant to squash them.  Stay tuned.

gdamore:
Rebased as 9b886a9999247d87c9f6d389c3e65a4bd39be010 (and squashed).

PR #307, opened 2018-03-21
fixes #306 Linger should be passed down to TCP sockets.

### LINKED ISSUE #304: perf tool exits too quickly on throughput client
The delay waiting for pipes to drain is insufficient, and the client does not include TCP lingering.

Noted and suggested by @skyformat99

// Wait 100msec for pipes to drain.
nng_msleep(100);      ====>   nng_msleep(1000);  

This is a band-aid, and we should do that, but we should also set the linger timeout.  It *should* in theory be the case that lingering is *sufficient*, but the real world knows that not all transports support a functional lingering, and I believe even our TCP transport implementation fails to set TCP_LINGER.

### LINKED ISSUE #306: Linger should be passed down to TCP sockets.
TCP should have the SO_LINGER option set.

Note that on macOS this should use SO_LINGER_SEC because SO_LINGER is counted in ticks.
(Arguably we could choose a finer grained linger, but that means determining the tick count.  It is also dependent upon the compile mode, so just using SO_LINGER_SEC is safer & easier if it is defined.)

Probably we should set this to 1 by default as well.

gdamore:
fixes #304 perf tool exits too quickly on throughput client

While here I also added most of the plumbing for TCP nodelay.

### DIFF
--- a/docs/man/nng_options.5.adoc
+++ b/docs/man/nng_options.5.adoc
@@ -81,10 +81,19 @@ This is the linger time of the socket in milliseconds.
 When this value is non-zero, then the system will
 attempt to defer closing until it has undelivered data, or until the specified
 timeout has expired.
+The actual time to linger will be rounded up based upon the granularity
+provided by the system; most often this is a whole second.
+This value defaults to one (with any rounding up applied.)
 
 NOTE: Not all transports support lingering, and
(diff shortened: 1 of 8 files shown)


gdamore:
This is wrong... see #308.

PR #345, opened 2018-04-14
fixes bad address tests in tcp and tls

toppk:
this will fix these two tests failing on my fedora 27 box

[toppk@static build]$  ../build/tests/tcp

Failures:

  * Assertion Failed (Malformed TCP addresses do not panic)
  File: ../tests/tcp.c
  Line: 54
  Test: nng_listen(s1, "tcp://127.0.0.1.32", ((void *)0), 0) == NNG_EADDRINVAL
  
FAIL    ../build/tests/tcp                                     1.093s
[toppk@static build]$  ../build/tests/tls

Failures:

  * Assertion Failed (Malformed TLS addresses do not panic)
  File: ../tests/tls.c
  Line: 259
  Test: nng_listen(s1, "tls+tcp://127.0.0.1.32", ((void *)0), 0) == NNG_EADDRINVAL
  
FAIL    ../build/tests/tls                                     2.531s
[toppk@static build]$

### DIFF
--- a/src/platform/posix/posix_resolv_gai.c
+++ b/src/platform/posix/posix_resolv_gai.c
@@ -98,6 +98,7 @@ nni_posix_gai_errno(int rv)
 		return (nni_plat_errno(errno));
 
 	case EAI_NONAME:
+	case EAI_NODATA:
 	case EAI_SERVICE:
 		return (NNG_EADDRINVAL);
 
--- a/src/platform/posix/posix_resolv_gai.c
+++ b/src/platform/posix/posix_resolv_gai.c
@@ -98,7 +98,9 @@ nni_posix_gai_errno(int rv)
 		return (nni_plat_errno(errno));
 
 	case EAI_NONAME:
+#ifdef EAI_NODATA
 	case EAI_NODATA:
+#endif
 	case EAI_SERVICE:
 		return (NNG_EADDRINVAL);
 


gdamore:
Cute.  This is not standard, it would appear, as this error value is not defined by POSIX.

http://pubs.opengroup.org/onlinepubs/007904875/functions/gai_strerror.html

So this value may not be defined/declared everywhere.  On macOS it is present, but protected behind a macro that prevents it from being exposed if _POSIX_C_SOURCE is defined.

See also this one:  https://lists.freebsd.org/pipermail/freebsd-ports/2003-October/005757.html

So I'm going to request that you add a check for #ifdef this value, to ensure that we always compile correctly.

src/platform/posix/posix_resolv_gai.c
```diff
@@ -98,6 +98,7 @@ nni_posix_gai_errno(int rv)
 		return (nni_plat_errno(errno));
 
 	case EAI_NONAME:
+	case EAI_NODATA:
```
gdamore: This value, EA_NODATA, is not present everywhere, and is not part of the POSIX standard.  Therefore, it's use here needs to be guarded by an #ifdef EAI_NODATA macro check.

toppk: Done as requested.

PR #386, opened 2018-04-30
fixes #385 Need NN_IPC_SEC_ATTR

### LINKED ISSUE #385: Need NN_IPC_SEC_ATTR
For legacy compat, we need to support the NN_IPC_SEC_ATTR option.  Note that this socket option was never documented.

gdamore:

contributing guidelines.


gdamore:
In retrospect, this idea was not good.  The test was borked because the original nanomsg option was borked.  The code assumed that it got to run before setting the option, and validation afterwards.  It could have set the bInheritHandle member, which was also busted.  An impedance mismatch here means we are better off without it.

Furthermore, this option was never documented.

PR #408, opened 2018-05-06
Portev

gdamore:
This has numerous fixes for building on illumos.  It includes a port-event based poller, with a new design that we will probably backport to the kqueue and epoll pollers (this design reduces overhead by avoiding the idhash lookups altogether.  It also minimizes lock contention on the master lock for the pollers.)


gdamore:
Replaced by #409

PR #409, opened 2018-05-06
Illumos2

gdamore:
This PR provides full illumos support for NNG.  The port event implementation is left as a separate commit, to facilitate testing apparent defects in the illumos epoll implementation.


gdamore:
Note that with these changes, "make test" comes back clean on omnios.  The make itself is also free from warnings.

PR #418, opened 2018-05-08
Fix double unlock in nni_posix_pollq_fini

AustinWise:


AustinWise:
I assume the intention was to reacquire the lock after waiting, not release the lock twice. I can create a ticket and update the commit message with it if you would like, however that seemed overkill for deleting two characters.

gdamore:
Good catch.  Thanks.

PR #420, opened 2018-05-08
fixes #419 want to nni_aio_stop without blocking

### LINKED ISSUE #419: want to nni_aio_stop without blocking
In general, we often do not need to block when we are stopping things down.  We would like to shut down all activity on a pipe, ensuring that nothing can be restarted on it, without necessarily needing to wait for callbacks to complete.

To this end, nni_aio_stop() could be modified to take a boolean, indicating whether to wait or not.

Of course, nni_aio_fini() must wait, and anything that is going to ensure that it is safe to free underlying resources must also wait.

gdamore:
This actually introduces an nni_aio_close() API that causes
nni_aio_begin to return NNG_ECLOSED, while scheduling a callback
on the AIO to do an NNG_ECLOSED as well.  This should be called
in non-blocking close() contexts instead of nni_aio_stop(), and
the cases where we call nni_aio_fini() multiple times are updated
updated to add nni_aio_stop() calls on all "interlinked" aios before
finalizing them.

Furthermore, we call nni_aio_close() as soon as practical in the
close path.  This closes an annoying race condition where the
callback from a lower subsystem could wind up rescheduling an
operation that we wanted to abort.

contributing guidelines.

### DIFF
--- a/src/core/aio.c
+++ b/src/core/aio.c
@@ -63,14 +63,15 @@ struct nng_aio {
 
 	// These fields are private to the aio framework.
 	nni_cv   a_cv;
-	bool     a_fini : 1;     // shutting down (no new operations)
-	bool     a_done : 1;     // operation has completed
-	bool     a_pend : 1;     // completion routine pending
-	bool     a_active : 1;   // aio was started
-	bool     a_expiring : 1; // expiration callback in progress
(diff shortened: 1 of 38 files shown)


gdamore:
Equivalent, yet superior, code was integrated with 16b4c4019c7b7904de171c588ed8c72ca732d2cf

PR #556, opened 2018-06-28
nni_url_parse does not handle ipc and inproc specially

widdma:
A missing return in nni_url_parse causes the special handling of ipc and inproc to be overwritten and the URL to be parsed completely. 

This results in the memory allocated by the strdup to leak.

It also means ipc and inproc URLs are canonicalized, not a bad thing per se, but this contradicts the documentation and breaks compatibility with nanomsg legacy code.


gdamore:
Nice catch.  The test failures appear to be unrelated, so I'll be merging this.

PR #635, opened 2018-08-09
fixes #634 test/tls and test/wssfile failures

### LINKED ISSUE #634: tests/tls and tests/wssfile fail the "Valid verify works" test in v1.0.0+
This is a failed attempt to build nng (tag v1.0.1) with TLS support against static mbedtls (tag mbedtls-2.12.0).

First, the relevant details:

Platform (Fedora 27):
```
$ uname -a
Linux ******** 4.17.11-100.fc27.x86_64 #1 SMP Mon Jul 30 15:22:33 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

NNG version (1.0.1):
```
$ cd $HOME/.local/src/nng
$ git status
HEAD detached at v1.0.1
```

mbedTLS version (2.12.0):
```
$ cd $HOME/.local/src/mbedtls
$ git status
HEAD detached at mbedtls-2.12.0
```

Both builds complete successfully (I used CMake).  The mbedTLS test suite completes successfully (100% pass).  The nng test suite fails the _tests/tls_ and _tests/wssfile_ tests.

I repeated the nng build using both the master branch and the v1.0.0 tag; same result (_tests/tls_ and _tests/wssfile_ fail).

I then assumed that this might be due to a change in mbedTLS, so I attempted to build and test nng v1.0.1 against mbedtls-2.11.0, mbedtls-2.7.4 and mbedtls-2.1.13 (the three mbedtls releases that most recently predate nng's 05-Jul-2018 1.0.1 release). Again, the same result (all mbedtls tests pass; nng's _tests/tls_ and _tests/wssfile_ fail).

Here are the specific failures:

```
$ ./tests/tls
Failures:

  * Assertion Failed (Valid verify works)
  File: $HOME/.local/src/nng/tests/tls.c
  Line: 267
  Test: nng_dialer_start(d, 0) == 0

FAIL    ./tests/tls                                            1.478s
```

```
$ ./tests/wssfile

Failures:

  * Assertion Failed (Valid verify works)
  File: $HOME/.local/src/nng/tests/wssfile.c
  Line: 269
  Test: nng_dialer_start(d, 0) == 0
 
FAIL    ./tests/wssfile                                        1.886s
```

mzipay:
This allows the tests to pass, but I think it likely that additional test cases and changes are needed; possibly in nng's wildcard domain processing and/or the same on mbedtls's side.

### DIFF
--- a/tests/tls.c
+++ b/tests/tls.c
@@ -448,7 +448,7 @@ TestMain("TLS Transport", {
 			nng_close(s2);
 			nng_close(s1);
 		});
-		trantest_next_address(addr, "tls+tcp://*:%u");
+		trantest_next_address(addr, "tls+tcp://localhost:%u");
 		So(nng_listener_create(&l, s1, addr) == 0);
 		So(init_listener_tls_file(l) == 0);
 		So(nng_listener_start(l, 0) == 0);
--- a/tests/wssfile.c
+++ b/tests/wssfile.c
@@ -371,7 +371,7 @@ TestMain("WebSocket Secure (TLS) Transport (file based)", {
 			nng_close(s2);
 			nng_close(s1);
 		});
-		trantest_next_address(addr, "wss://:%u/test");
+		trantest_next_address(addr, "wss://localhost:%u/test");
 		So(nng_listener_create(&l, s1, addr) == 0);
 		So(init_listener_wss_file(l) == 0);
 		So(nng_listener_start(l, 0) == 0);


gdamore:
Thanks for looking into this, I will be doing further research into it.

mzipay:
If it helps to spur further research... I think it might be necessary to add SANs for both localhost and 127.0.0.1 to the certificate in addition to CN.

ProfFan:
Probably you can just use a script to automatically acquire a free valid certificate from Let's Encrypt.

mzipay:
@ProfFan That wouldn't even be necessary. A self-signed certificate with SANs accounting for "localhost" and "127.0.01" would be sufficient. I'd just use openssl for that, no need for a public CA.

The issue is the ability to fully exercise the test suite w/r/t wildcard hosts; even my patch doesn't accomplish that (it only exercises the case where the host is explicit, but wildcard hosts are also a valid case).

It might be argued that the test suite could be left with only my patched change given that (a) support for wildcard hosts is tested elsewhere and (b) the tests that are failing are not actually due to any bug in nng but rather are a result of how the test cert is generated... assuming my analysis is correct ;)

gdamore:
I will get to looking at this soon.  I just haven’t had time yet.  I may
take a look at it on the plane tomorrow (I have a very long flight as I’m
coming from a two week trip to Russia.)

gdamore:
This does not feel entirely right to me.  The cert has a CN of localhost, which should be adequate.  What version of mbedTLS are you using?

mzipay:
I am using mbedTLS 2.12.0 and nng 1.0.1. (All version info for both nng and mbedTLS, including other variations/combinations that I tried, are detailed in issue #634.)

I don't think the issue is with the cert per se, but rather with how the test establishes the connection, given that the cert is issued _only_ for CN=localhost.

The tests/tls.c and tests/wssfile.c tests that fail ("Valid verify works") are not connecting (explicitly) to the host named in cert's CN ("localhost"), and this is I believe what is causing the issue, given that both tests pass if the host is named explicitly in the connection.

My understanding is that to support "wildcard" hosts, a cert needs to be generated with both a CN and SANs (subjectAltName).  So for passing tests, either the "Valid verify works" tests would need to connect using a host that matches the cert CN, or the cert would need to include SANs as well.

If you attempt to build nng 1.0.1 with mbedTLS 2.12.0, do tests/tls.c and tests/wssfile.c pass or fail the "Valid verify works" tests?

I agree that it doesn't quite feel right. I'd expect that connecting using a wildcard host would work; a wildcard should accommodate "CN=localhost" (in my mind, at least). But it doesn't seem to be the case.

gdamore:
So as I understand it, the problem is that we have two different forms of binding, and there is probably also a difference in the handling of AF_UNSPEC by your platform.  This is fairly typical for some systems -- the * wildcard should in theory bind to both IPv6 and IPv4 networks, but maybe isn't doing that on your system (can depend on socket options).

A better solution, imo, is to force IPv4 for both sides by changing the URLs to use tls+tcp4://  (note the "4" at the end).  This is a relatively newer feature in NNG, but it constrains the socket to use IPv4 exclusively.  (There's a similar feature for tls+tcp6:// btw.)

I'm going to close this PR for now.  If you want to submit a change that only changes the scheme, and leaves the rest unaltered (provided that actually works for you), I'd be happy to approve *that*.

mzipay:
Sure, that sounds like a good approach for the tls test. Do you have any thoughts on the wss test? Does it support wss4/wss6 alternative syntax too? Apologies if that's a naive question; I don't have any experience working with WebSockets.

PR #657, opened 2018-08-19
fixed compilation error on OpenBSD, missing sockpeercred added as nng…

opunix:
@gdamore as you requested in Gitter :-) ... here is the fix for the compilation error on OpenBSD

BR Frank


opunix:
ohh there is a line which needs to be fixed sorry for that

PR #697, opened 2018-09-02
Make docs for old versions and tip point to the correct page

codypiersall:
All the man page docs that are not the latest released version give a notice, saying that the current docs being viewed are not the latest, and offer a link to the latest.  This fixes the broken link; formerly it was pointing to nng.html, now it points to index.html.


codypiersall:
If you'd rather close this PR and just make the change yourself, feel free.  It doesn't make a difference to me.

gdamore:
Thanks for fixing this.

PR #725, opened 2018-09-15
Only free in nng_send() if we succeed.

lkesteloot:
Both the man page and the comment (in nng.h) of nng_send()
say that the buffer is freed only if the function succeeds.
The previous code always freed (if the flag was set). This
change only frees if we succeed.


gdamore:
Good catch.

PR #742, opened 2018-10-07
Small fix

QXSoftware:
Yet another small fix

### DIFF
--- a/src/supplemental/util/platform.c
+++ b/src/supplemental/util/platform.c
@@ -44,7 +44,6 @@ nng_thread_create(nng_thread **thrp, void (*func)(void *), void *arg)
 	if ((thr = NNI_ALLOC_STRUCT(thr)) == NULL) {
 		return (NNG_ENOMEM);
 	}
-	memset(thr, 0, sizeof(*thr));
 	*thrp = (void *) thr;
 	if ((rv = nni_thr_init(thr, func, arg)) != 0) {
 		return (rv);
--- a/docs/man/nng_req_open.3.adoc
+++ b/docs/man/nng_req_open.3.adoc
@@ -11,7 +11,7 @@
 
 == NAME
 
-nng_req_open - create rep socket
+nng_req_open - create req socket
 
 == SYNOPSIS
 
--- a/src/core/message.c
+++ b/src/core/message.c
@@ -445,7 +445,6 @@ nni_msg_dup(nni_msg **dup, const nni_msg *src)
 	if ((m = NNI_ALLOC_STRUCT(m)) == NULL) {
 		return (NNG_ENOMEM);
 	}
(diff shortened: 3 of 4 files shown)


src/core/timer.c
```diff
@@ -37,7 +37,6 @@ nni_timer_sys_init(void)
 	int        rv;
 	nni_timer *timer = &nni_global_timer;
 
-	memset(timer, 0, sizeof(*timer));
```
gdamore: I think this one is still necessary.  We don't clear t_waiting in the section below, and there is a very low chance that a user can call nng_fini() causing a teardown, and then subsequently cause a new initialization.  This is not on any hot code path (library initialization only), so removing it doesn't really save anything.

src/supplemental/util/platform.c
```diff
@@ -44,7 +44,6 @@ nng_thread_create(nng_thread **thrp, void (*func)(void *), void *arg)
 	if ((thr = NNI_ALLOC_STRUCT(thr)) == NULL) {
 		return (NNG_ENOMEM);
 	}
-	memset(thr, 0, sizeof(*thr));
```
gdamore: Another leftover... :-) Originally NNI_ALLOC_STRUCT() didn't promise to clear memory like it does now.

gdamore:
One of those memset()'s still need to be in there.  The rest looks fine.

src/core/timer.c
QXSoftware: Hi~nni_global_timer

QXSoftware: Hi~
The **nni_global_timer** is a static variable, so it will be zero by default isn't it?

src/supplemental/util/platform.c
QXSoftware: Yeah, I searched the whole project with keyword **NNI_ALLOC_STRUCT** and found two residuals~

src/core/timer.c
QXSoftware: After a few days finally I got what you mean, the library is capable of closing/opening multiple times. So the `memset` in `nni_timer_sys_init` is still necessary. A new commit had already been pushed.

QXSoftware:
Hi~

A new commit had been pushed, adding the `memset` back in `nni_timer_sys_init`. Please review, thank you!

gdamore:
This all looks good now, but can you resubmit with your last commit squashed with the first one?  (I.e. your last commit is a correction to the first one, so let's not leave the mistake in the git history...)

QXSoftware:
> This all looks good now, but can you resubmit with your last commit squashed with the first one? (I.e. your last commit is a correction to the first one, so let's not leave the mistake in the git history...)

Sure! It's a very good habit!

PR #747, opened 2018-10-09
Allow recv_maxsize to be set after calling listen() on listener.

### LINKED ISSUE #724: `NNG_RECVMAXSZ` option can be disregarded if `nng_listen()` happens before `nng_setopt(...)`.
our contribution guidelines.

## NNG & Platform details.

Current nng master branch, Windows.  I ran into this because I had an [Appveyor build fail on Windows](https://ci.appveyor.com/project/codypiersall/pynng/build/1.0.41), but my [Travis CI runs were fine](https://travis-ci.org/codypiersall/pynng/builds/427502261).

Importantly, I ran a `git bisect` against my unit tests, and I was able to narrow down the breaking changes to commit 5b61fa0e04348ee7661dbb19bc9c50ca22af74dc, and I wrote a C program that demonstrates the problem below (so the bindings that I wrote are not involved at all, as you'll see below).

## Expected Behavior

Order of `nng_listen()` and `nng_setopt_size(sock, NNG_RECVMAXSZ, size)` shouldn't matter

## Actual Behavior

Order appears to matter (on Windows! I didn't actually test this code on Linux, but I _suspect_ that Linux is fine since the bindings I'm writing work on Linux.).

## Steps to Reproduce

Run the program below twice: once with arguments `program.exe listen_first`, and once as `program.exe listen_last`.

```c
#include <assert.h>
#include <stdbool.h>
#include <stdio.h>
#include <string.h>
#include "nng.h"
#include "protocol/pair1/pair.h"

int main(int argc, char *argv[]) {
	nng_socket a;
	nng_socket b;
	nng_listener l;
	// don't actually care about the contents of the message
	bool listen_first = true;
	if (argc >= 2) {
		if (strcmp("listen_first", argv[1]) == 0) {
			listen_first = true;
		}
		else if (strcmp("listen_last", argv[1]) == 0) {
			listen_first = false;
		}
	}

	char sendbuf[100];
	char recvbuf[100];
	char addr[] = "tcp://127.0.0.1:31313";
	size_t recvbufsize = 100;

	assert(nng_pair1_open(&a) == 0);
	if (listen_first) {
		printf("Listening first fails\n");
		assert(nng_listen(a, addr, &l, 0) == 0);
		assert(nng_setopt_size(a, NNG_OPT_RECVMAXSZ, 100) == 0);
	}
	else {
		printf("Listening after works\n");
		assert(nng_setopt_size(a, NNG_OPT_RECVMAXSZ, 100) == 0);
		assert(nng_listen(a, addr, &l, 0) == 0);
	}
	size_t listener_maxsize;
	size_t sock_maxsize;
        // They socket and listener are actually reporting the correct size...
	assert(nng_listener_getopt_size(l, NNG_OPT_RECVMAXSZ, &listener_maxsize) == 0);
	assert(nng_getopt_size(a, NNG_OPT_RECVMAXSZ, &sock_maxsize) == 0);
	assert(listener_maxsize == 100 && sock_maxsize == 100)

	assert(nng_setopt_ms(a, NNG_OPT_RECVTIMEO, 100) == 0);
	assert(nng_pair1_open(&b) == 0);
	assert(nng_dial(b, addr, NULL, 0) == 0);

	assert(nng_send(b, sendbuf, 100, 0) == 0);
	if (nng_recv(a, recvbuf, &recvbufsize, 0) == NNG_ETIMEDOUT) {
		printf("It worked!\n");
	}
	else {
		printf("Sad times, didn't work!\n");
	}
	assert(nng_close(a) == 0);
	assert(nng_close(b) == 0);
	return 0;
}
```

Thanks for all the work you're putting into nng!

codypiersall:
fixes #724 recv_maxsize is ignored if nng_listen() called before set.

This also adds a test to ensure that the fix works.  

I have a couple questions about my PR:

1. I'm not holding the pipe mutex while changing the struct members, since I didn't think the mutex was necessary; it doesn't really matter if the struct member changes.  I guessed that the endpoint mutex would protect against any pipes getting deleted, so I thought this wouldn't crash anything.
2. On the test, I'm testing ipc, inproc, tcp, and ws.  Are any other tests needed?  And I think I probably need to check that those transports are supported instead of assuming that they are?
3. incproc and websocket didn't require any changes; they just worked.
4. I didn't test zerotier, but I can (probably?) figure out how to test it if that's required.
5. I just chose a couple random ports for tcp and ws, which may not be great?

Primarily, this allows overriding recv_maxsize on a per-listener basis more naturally (to me, anyway), since code written like this will now just work:

``` c
nng_socket s;
nng_listener l1;
nng_listener l2;
nng_pair1_open(&s);
// listen on local interface and public
nng_listen(s, "tcp://127.0.0.1:11111", &l1, 0);
nng_listener_setopt_size(l1, NNG_OPT_RECVMAXSZ, -1);  // unlimited on localhost
nng_listen(s, "tcp://X.X.X.X:11111", &l2, 0);
nng_listener_setopt_size(l2, NNG_OPT_RECVMAXSZ, 1024 * 1024);  // 1 MB external
```


codypiersall:
Ah, interesting – I didn't realize that pipes were meant to be immutable after creation.  I think there are some inconsistencies here, though; for some transports, changing the rcvmaxsz on the listener works after listen, and for some it doesn't.  Specifically, **even without this patch, inproc and websocket still actually allow setting recv max size after the endpoint is started**; my patch isn't required.  I'm not sure about zerotier.

My preference would be that the transports all act the same, if possible; one of the great for me about nng is that I don't really have to think about the details of the transport I'm using, and if I need to change to a different transport it has traditionally only required me to change one line of my code; the address that is being connected to/received from.  The fewer details that are different between the different transports, the better.  Does that make sense?

I can think of a couple ways to resolve this, one way or another:

1. If setting the recv max size after listener/dialer start is acceptable, but changing the field on the pipes for tcp transport and ipc transport is not, would it be acceptable to look up the listener value on the endpoint associated with the pipe?  As far as I can tell, the recv max size is only used when a new message comes in, so it shouldn't be on the fast path; anyway, it's not a lot of work to look it up on the endpoint anyway. This would be my preference.
2. If setting the recv max size is not supported after the endpoint is started, my preference would be to disallow it on the other transports as well after they are started.

What's the rationale for not allowing to change this attribute after the endpoint is started?  I understand that with certain attributes there may be transport-level restrictions; e.g. if you change a socket parameter in one thread while the socket is doing things in another, there could easily be some TOCTOU bugs.  Since this attribute exists at a higher level, though, it would just mean either accepting a message or rejecting it if there was a race.  As far as I can tell, anyway.

It's also worth mentioning that this is new behavior that was introduced in 5b61fa0e04348ee7661dbb19bc9c50ca22af74dc; before that, you could set recv max size before or after an endpoint was connected.

Anyway, what do you think?  I'd be willing to do the work for either option 1. or 2. above if you think either is better than the status quo.

gdamore:
I think inproc doesn't honor the recv size at all right now.  You can set it however you like I think.

The websocket thing is a consequence of the websocket implementation, and not something folks should depend upon.

I am thinking number 2 is the way to go.

The rationale for not supporting dynamic changes is that it might not always be *possible* to support this.  It is for the current transports we have now, but that might change in the future if transports negotiate for a common maximum size for example.  (And there are some reasons to do just this -- so that we can send an error back to the sending user instead of dropping his message on the floor.)

What I'm thinking is that it would be better to just have folks stop assuming these values are mutable after dial/listen (or endpoint start).  And not just these, but pretty much any others.  We could actually return a sensible error message back to them (NNG_EBUSY or somesuch).

At least that would lead to a consistent and predictable semantic.

gdamore:
The other reason to lock this down is that prevents a whole bunch of potential race conditions.  Since this option exists mostly for the purposes of security (preventing denial of service), you really *do* want that option set *before* a connection is made and someone starts sending you bad data.

gdamore:
Interestingly enough, I think this build failure would be fixed by this PR:

https://circleci.com/gh/nanomsg/nng/2718?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification

I really need to figure out what I want to do here.

codypiersall:
Ah yeah, I think you're right about that failure.  If anyone else is following this discussion, I'll lay out the reason that that test sometimes-fails: it's the [compat_tcp](https://github.com/nanomsg/nng/blob/master/tests/compat_tcp.c#L206) test, starting around line 206.  In that test, the listener is started with the `nn_bind()` function before recv max size is set.  However, since it is using nn_bind, where the actual binding happens in the background; therefore there is a race between the listener _actually_ starting and the option getting set.  Most of the time, apparently, the option gets set before the listen starts.  Please correct any of that if I was wrong; I may have gotten some of the _details_ wrong, but the race part I think I got right.

I'd still prefer to be able to change the option _after_ the listen call, but I agree with you that it does create a race, and with potential security implications.  I don't think that it's a very likely attack avenue, though :-).

I started looking into implementing the number 2 option above, where it would be strictly not allowed to set `NNG_OPT_RECVMAXSZ` after an endpoint is started, but unfortunately is was slightly more complicated than I had hoped to implement it the way I wanted.  My though was to special-case that option inside of [`nni_listener_setopt`](https://github.com/nanomsg/nng/blob/156604bd07ee60faa323968c71627f1c701b473a/src/core/listener.c#L364), so it would send `ESTATE` whenever the listener was already started.  The reasoning was that it would only need to be implemented there instead of in each of the transport protocols.  However, the struct field [`nni_listener.l_started`](https://github.com/nanomsg/nng/blob/master/src/core/sockimpl.h#L87) is an `atomic_flag`, and so its state can't be queried.  Does that field mean what I think it means?  It looks like it wouldn't really be a performance hit to turn it into an atomic bool, but I wasn't looking forward to figuring out what the equivalent was on Windows, and I don't have a way of testing on a Mac :-/.

It's been a bit of a crazy few weeks personally for me, so I haven't been able to look into this as much as I'd have liked; my almost-5-month old seems to be coming into a better sleep pattern though, so hopefully I can devote more time to side projects.

gdamore:
I'm slowly coming around to the thinking that we should enable this, because -- yeah, compatibility protocols kind of need this to work.

My guess is that we need to iterate over all the pipes, and somewhere we need to carefully guard against race conditions during pipe attach.  Stay tuned.

gdamore:
Just went back over this - and I think this is about right.

We need to make similar logic changes for the other transports though.

codypiersall:
I just rebased my commit to work with the new `include` layout.

The [test that I added](https://github.com/nanomsg/nng/blob/03d7a6a6cf409bd7bb02fdad7c6187c3b7f44786/tests/set_recvmaxsize.c) has tests for inproc, ipc, tcp, and ws and they all pass.  Which other transports need to be changed?

PR #754, opened 2018-10-16
fix #751

### LINKED ISSUE #751: http server directory handler
The directory handler created via `nng_http_handler_alloc_directory(nng_http_handler **hp, const char *path, const char *dirname);` cannot accept uri "/" (the `path` parameter), causing 404 error. Is this behaviour by design or a mistake?

QXSoftware:
fixes #751

### DIFF
--- a/src/supplemental/http/http_server.c
+++ b/src/supplemental/http/http_server.c
@@ -560,17 +560,21 @@ http_sconn_rxdone(void *arg)
 		if (strncmp(path, h->uri, len) != 0) {
 			continue;
 		}
-		switch (path[len]) {
-		case '\0':
-			break;
-		case '/':
-			if ((path[len + 1] != '\0') && (!h->tree)) {
-				// Trailing component and not a directory.
-				continue;
+		if (len == 1 && h->uri[0] == '/') {
+			// deal with root uri "/"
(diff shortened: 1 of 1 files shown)


src/supplemental/http/http_server.c
```diff
@@ -560,17 +560,21 @@ http_sconn_rxdone(void *arg)
...
 			continue;
 		}
-		switch (path[len]) {
-		case '\0':
-			break;
-		case '/':
-			if ((path[len + 1] != '\0') && (!h->tree)) {
-				// Trailing component and not a directory.
-				continue;
+		if (len == 1 && h->uri[0] == '/') {
```
gdamore: This code is logically not right.

You're bypassing the entire tree check here for one.

What is the value of h->uri here and path here?

QXSoftware:
For now register on "" is the same as "/", take a look at [nni_http_handler_init](https://github.com/nanomsg/nng/blob/master/src/supplemental/http/http_server.c#L99-L101).

gdamore:
Thanks for that reference.  I now wonder if I got this wrong.  Maybe instead of turning "" into "/" it should be the other way around, turning "/" into "".

PR #756, opened 2018-10-17
fixes warning for comparing signed and unsigned

janjaapbos:
fixes compile warning

### DIFF
--- a/src/transport/zerotier/zerotier.c
+++ b/src/transport/zerotier/zerotier.c
@@ -845,7 +845,7 @@ zt_pipe_recv_data(zt_pipe *p, const uint8_t *data, size_t len)
 	// that we only can catch the case where a message is larger by
 	// more than a fragment, since the final fragment may be shorter,
 	// and we won't know that until we receive it.
-	if ((nfrags * fragsz) >= (p->zp_rcvmax + fragsz)) {
+	if ((size_t)(nfrags * fragsz) >= ((p->zp_rcvmax) + fragsz)) {
 		// Discard, as the forwarder might be on the other side
 		// of a device. This is gentler than just shutting the pipe
 		// down.  Sending a remote error might be polite, but since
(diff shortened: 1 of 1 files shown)


src/transport/zerotier/zerotier.c
```diff
@@ -1778,7 +1778,7 @@ zt_pipe_send(void *arg, nni_aio *aio)
 	fragsz = (uint16_t)(p->zp_mtu - zt_offset_data_data);
 
 	bytes = nni_msg_header_len(m) + nni_msg_len(m);
-	if (bytes >= (0xfffe * fragsz)) {
+	if ((uint16_t)bytes >= (0xfffe * fragsz)) {
```
gdamore: This needs to be size_t I think.  Or possibly just "unsigned".  The thing is that bytes might overflow a 16-bit value, which is why this check is here.

src/transport/zerotier/zerotier.c
```diff
@@ -845,7 +845,7 @@ zt_pipe_recv_data(zt_pipe *p, const uint8_t *data, size_t len)
 	// that we only can catch the case where a message is larger by
 	// more than a fragment, since the final fragment may be shorter,
 	// and we won't know that until we receive it.
-	if ((nfrags * fragsz) >= (p->zp_rcvmax + fragsz)) {
+	if ((nfrags * fragsz) >= ((uint16_t)(p->zp_rcvmax) + fragsz)) {
```
gdamore: Yeah, pretty sure uint16_t is too small here.  This needs to compare against uintptr_t or something like that.  (I suspect size_t is right, actually.)

Or (and I think this is better) -- cast the *other* side to int.  I think we want to cast "up" rather than down.

src/transport/zerotier/zerotier.c
gdamore: Yeah, pretty sure we want to cast to (int) on the other side.  Because fragsz is the more constrained value.

src/transport/zerotier/zerotier.c
janjaapbos: Thanks! Yes, I found out now that nothing was coming through. ;-)

janjaapbos:
I think this ok to merge?

gdamore:
Sorry, got side tracked, and this fell through the cracks.  While your change no doubt silences the warnings, I'm worried about overflow errors that can occur, and I don't think your change addresses that -- because it casts *after* the multiplication rather than before.

I think what we should do is change the actual data type for nfrags etc. to size_t.  I was looking at that briefly the other day, but as I said I got side tracked before I was able to finish following through.

I'll fix this today, but I'll probably not use this PR as is, because of those overflow concerns.

gdamore:
This was fixed a different way, see b1b918378513566a9359e655ea538292c62d4173

PR #790, opened 2018-11-14
Fix thread crash on websocket ping

### LINKED ISSUE #789: Websocket ping request deadlock
Thanks @gdamore for this excellent lib.

I found a problem with websockets, when running on Windows.
When the client does a ping request, the server locks twice with AcquireSRWLockExclusive and ends up in a deadlock, crashing the thread.

I couldn't find proper documentation on AcquireSRWLockExclusive, but this answer on SO says that re-acquiring is not possible: https://stackoverflow.com/a/30708100

ghost:
fixes #789  Websocket ping request crash

I wish I could add a unit test but sending PING opcode is currently not possible.

### DIFF
--- a/src/supplemental/websocket/websocket.c
+++ b/src/supplemental/websocket/websocket.c
@@ -689,7 +689,6 @@ ws_send_control(nni_ws *ws, uint8_t op, uint8_t *buf, size_t len)
 
 	if ((ws->closed) ||
 	    (ws_msg_init_control(&wm, ws, op, buf, len) != 0)) {
-		nni_mtx_unlock(&ws->mtx);
 		return;
 	}
 


gdamore:
Thanks for looking at this.  I would like tl take some time to look over
this to ensure that this is correct and we either do not need the lock here
or are always in a context where the caller already holds the lock.

gdamore:
I see only one call site, so this is probably correct, but the changeset needs to remove the unlock from the error path in the same function.

src/supplemental/websocket/websocket.c
```diff
@@ -687,7 +687,6 @@ ws_send_control(nni_ws *ws, uint8_t op, uint8_t *buf, size_t len)
 
 	// Note that we do not care if this works or not.  So no AIO needed.
 
-	nni_mtx_lock(&ws->mtx);
 	if ((ws->closed) ||
 	    (ws_msg_init_control(&wm, ws, op, buf, len) != 0)) {
 		nni_mtx_unlock(&ws->mtx);
```
gdamore: You need to remove this unlock too.

gdamore:
If you want me to just fix this myself, let me know... the changes needed are fairly obvious, and I'd like to get this into a release quickly if possible.  Thanks.

src/supplemental/websocket/websocket.c
bertrand-: Oupsy... correct

bertrand-:
Yes please change it yourself, it is very straightforward. I will drop the PR.

Thanks a lot Garrett

bertrand-:
Just pushed the removal of the unlock, if it can help.

Thanks a lot

PR #795, opened 2018-11-19
Allow nng to be consumed in "add_subdirectory" scenarios

gregorburger:
This change allows to use nng as part of a bigger project using add_subdirectory.

e.g.: somebody adds nng as external dependency using:
```
FetchContent_Declare(
    nng
    URL https://github.com/nanomsg/nng/archive/v1.1.0.zip
)

FetchContent_GetProperties(nng)
if (NOT nng_POPULATED)
    FetchContent_Populate(nng)
    add_subdirectory(${nng_SOURCE_DIR} ${nng_BINARY_DIR})
endif()
```
and then later uses nng as target link library:
`target_link_library(nng_example nng)`

it will find the headers of (e.g. nng.h) using this change

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -246,7 +246,7 @@ set_target_properties (${PROJECT_NAME} ${PROJECT_NAME}
 
 target_link_libraries (${PROJECT_NAME} PRIVATE ${NNG_LIBS})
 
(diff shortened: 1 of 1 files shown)


gdamore:
Thanks for this.  Earlier this past weekend I think I discovered the same deficiency.  Glad you fixed it!

PR #796, opened 2018-11-20
allow to use nng in combination with other projects depending on nng in add_subdirectory scenarios

gregorburger:
this allows to use nng in combination with nngpp as cmake external project dependency
external projects require that nng.h is in nng folder, which is not the case in the source layout


gdamore:
I didn't realize that the last PR you submitted was insufficient.  I want to hold this right now, because I want to get a patch release (1.1.1) out soon.  Once I've done that, I'll consider this for integration.

gdamore:
I think there is probably a better to do this, rather than reorganizing the entire tree, as well.  The BUILD_INTERFACE approach feels like the right approach, combined probably with an export() to export values into parent projects.  I've had good luck with that in another project.

gregorburger:
The first pull request made it possible to use nng inside other cmake projects IF you use nng directly. You would also need to use `#include <nng.h>` without being inside the folder nng because they still reside inside the src folder. BUT when you use nng indirectly via for example nngpp then it does not work because nngpp expects the header `nng.h` inside of the folder nng so it does a `#include <nng/nng.h>` instead of `#include <nng.h>`. This problem will also come up if you switch from embedded nng to system installed nng. 

A less intrusive change would be rename the `src` folder to `nng` folder. Then you could change the BUILD_INTERFACE include directory to be the PROJECT_DIRECTORY and it will find the include in nng/nng.h.

gregorburger:
But having a `include` folder that hosts the public headers besides the `src` folder is something i see in a lot of projects that particularly support `add_subdirectory` scenarios. So it seems that this is the common best practice to deal with cmake projects which have public headers. That is why I chose the more "radical" approach.

gdamore:
I'm inclined to accept this change, in spite of the size of it.  I can see some particular benefit in separating "public" headers from internal ones... and that is probably main benefit that convinced me.

There is a trivial conflict to resolve in nng.h itself, can you please resolve and resubmit.  I'd also prefer these commits squashed (if you don't squash them, I will.)

gregorburger:
I pulled in your master branch. but i can't figure out how to change the pull request into a squashed one, but i saw somewhere that you can squash it during the merge, is that true?

gdamore:
yes i can. i will integrate it later.

gdamore:
So the way you can do this yourself in your own projects is to use git rebase -i from the command line.  That's an interactive rebase, but it lets you squash commits selectively, and also for example elide commits that are duplicates, etc.  You should play with that on your own.

gdamore:
I found a missed header change in the zerotier test... Will fix up.  Stay tuned.

gdamore:
I'm also sorting the include directives a bit -- the tests should use <> for including public headers now, I think, and I prefer to have the order be

```
#include <systemheader1.h>
#include <systemheader2.h>

#include <nng/nng.h>
#include <nng/transport/tcp/tcp.h>

#include "localheader.h"
```

(So standard headers first, followed by public NNG headers, followed by directory specific private headers, including includes that reach into other private directories.)

gdamore:
My updates are here: https://github.com/nanomsg/nng/pull/798/commits/9af760948bf7dd3bf5c4d9f795c6b26b352dd8cd

I will handle this in a subsidiary PR.

gdamore:
The changes originally proposed have been merged as part of https://github.com/nanomsg/nng/commit/d6bb25e1e0a25cb5aa781ac4f90b513fd5624f50

Closing this PR.  Thanks for your contribution.

PR #806, opened 2018-11-29
Only include http_public if http is enabled

zagor:
http_public.c is always compiled and included, even with NNG_ENABLE_HTTP=OFF
This patch changes that.


gdamore:
So, this change isn't approved.  The public HTTP API is still "present", but the functions are stripped down to simply return NNG_ENOTSUP.  This allows programs to still be linked, avoiding the need for consuming programs and libraries to inject '#ifdef' statements into their code.  Yes this slightly increases the size of the library, but it should only be by a few kilobytes, since all the "real" code is stripped out inside the http public function.

PR #810, opened 2018-12-08
The toplevel sets threads already, not needed here.

codypiersall:
I'm working on updating my Python bindings to use the new source code layout (splitting out `include`), and found out that nng no longer builds correctly on the Docker image I use to build packages on CI.  I [made an issue in pynng](https://github.com/codypiersall/pynng/issues/15) and I have tracked down the breaking change to https://github.com/nanomsg/nng/commit/77984b14a65c0a7387c97f3d36f947dd17358744.

To be honest I'm not really sure what's going on.  For some reason, in the Docker image, the test programs aren't getting linked with `librt` for `clock_gettime()`, but I'm not seeing the issue on my actual Linux box.

As far as I can tell, this change fixes the issues, but I really don't know why.  I don't know much about CMake.  There's probably a better fix :-).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -89,6 +89,7 @@ set (NNG_SRCS
 
 if (NNG_PLATFORM_POSIX)
     find_package (Threads REQUIRED)
+    list(APPEND NNG_LIBS Threads::Threads)
 
     set (NNG_SRCS ${NNG_SRCS}
         platform/posix/posix_impl.h
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -89,7 +89,6 @@ set (NNG_SRCS
 
 if (NNG_PLATFORM_POSIX)
     find_package (Threads REQUIRED)
-    set(NNG_LIBS Threads::Threads)
 
     set (NNG_SRCS ${NNG_SRCS}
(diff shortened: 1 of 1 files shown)


codypiersall:
Oops, I meant to link to my failing CI runs: https://travis-ci.org/codypiersall/pynng/builds/464143046#L697

It's complaining about not finding `clock_gettime()`.

gdamore:
Looking at this a bit more closely, I think you're on the right path, but not quite right.

Probably this should be a list append to NNG_LIBS (so we don't override whatever value was found before).  Removing it may work for you because it turns out that we have explicit checks for sem_wait() against pthread, but that's not the right approach, really.

src/CMakeLists.txt
```diff
@@ -89,7 +89,6 @@ set (NNG_SRCS
 
 if (NNG_PLATFORM_POSIX)
     find_package (Threads REQUIRED)
-    set(NNG_LIBS Threads::Threads)
```
gdamore: This should turn into an append, not be removed entirely.  (The removal is working because we also pick up on -lpthread elsewhere.)

codypiersall:
Thanks for the feedback, @gdamore.  Makes sense!  I've updated my branch.

PR #811, opened 2018-12-10
fixes #769 How to limit worker threads

### LINKED ISSUE #769: How to limit worker threads
Is there a way to limit the number of worker threads that NNG uses?

We have some execution environments with very limited memory resources.

We are trying to limit or reduce the number of our own threads, and this is hard
enough without NNG going in the opposite direction.

mjgigli:
Although there are a number of recommendations in #769 on how to reduce the memory footprint, I am submitting a PR that deals mostly with just the title: provide a mechanism for limiting the number of worker threads. This PR exposes CMake variables to set the number of resolver and taskq threads, respectively. Maybe a separate Issue is warranted, for tracking change to default stack sizes of various internal threads?

I tested this using the async demo on a Ubuntu VM running on Windows. By default, 25 threads are created on my machine for the async demo, and with these changes I was able to reduce the number of resolver threads to 1 and the number of taskq threads to 2:

```
$ cmake -G Ninja -DNNG_RESOLV_CONCURRENCY=1 -DNNG_NUM_TASKQ_THREADS=2 ..
```

With these changes, I was able to run the async demo with 8 total threads instead of 25.

Reducing the memory footprint of nng is really important to my team for using nng on an embedded target, and this was an easy way for us to move in the right direction.

### DIFF
--- a/src/platform/posix/posix_resolv_gai.c
+++ b/src/platform/posix/posix_resolv_gai.c
@@ -34,15 +34,15 @@
 // changed with this define.  Note that some platforms may not have a
 // thread-safe getaddrinfo().  In that case they should set this to 1.
 
-#ifndef NNG_POSIX_RESOLV_CONCURRENCY
-#define NNG_POSIX_RESOLV_CONCURRENCY 4
+#ifndef NNG_RESOLV_CONCURRENCY
+#define NNG_RESOLV_CONCURRENCY 4
 #endif
 
 static nni_mtx  resolv_mtx;
 static nni_cv   resolv_cv;
 static bool     resolv_fini;
 static nni_list resolv_aios;
-static nni_thr  resolv_thrs[NNG_POSIX_RESOLV_CONCURRENCY];
+static nni_thr  resolv_thrs[NNG_RESOLV_CONCURRENCY];
 
 typedef struct resolv_item resolv_item;
 struct resolv_item {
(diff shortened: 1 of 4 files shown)


src/platform/posix/posix_resolv_gai.c
```diff
@@ -34,6 +34,10 @@
 // changed with this define.  Note that some platforms may not have a
 // thread-safe getaddrinfo().  In that case they should set this to 1.
 
+#ifdef NNG_RESOLV_CONCURRENCY
+#define NNG_POSIX_RESOLV_CONCURRENCY NNG_RESOLV_CONCURRENCY
+#endif
+
 #ifndef NNG_POSIX_RESOLV_CONCURRENCY
```
gdamore: My gut instinct is that this can be made slightly better.

Basically, I think we can eliminate the NNG_POSIX_RESOLV_CONCURRENCY (and also the Windows one below) and replace this with NNG_RESOLV_CONCURRENCY:

```
#ifndef NNG_RESOLV_CONCURRENCY
#define NNG_RESOLV_CONCURRENCY 4
#endif
```

Then the references to the old values should be replaced with this one new value.

It's a small change, but I think that will result in cleaner and clearer code, in case someone comes back here in the future.

gdamore: Basically what I mean is, we should just rename the old values to the new value.  Then you don't need the new block of code, because the old code already takes care of it.  Just need to update for your new unified name.

mjgigli: Makes sense. I was trying not to change too much, but I agree it would be clearer to just replace the platform-specific macros. I'll make the change.

mjgigli:
Thanks for the review. I can make the change, should be able to get to it tonight.

mjgigli:
@gdamore updated to collapse the platform-specific resolve concurrency macros. Let me know if you'd like me to squash the commits.

Also, is there any documentation I should update to describe the two new CMake variables?

gdamore:
This looks really good.  I'm happy to squash these commits, though generally in an ideal world I prefer that submitters do so.  I'll wait a day, to give you a chance to do so; otherwise I'll do it myself.

PR #860, opened 2019-01-15
fixes typos in stat man pages

jake-ruyi:


gdamore:
Thanks for catching these!

PR #893, opened 2019-02-24
This makes sub use contexts.

gdamore:
It should be quite a lot lighter weight, although the code for
matching subscriptions is still quite suboptimal.  We will be
replacing that with a proper patricia trie later.

xsub still uses the old msgqueue implementation.


gdamore:
This code was integrated in the subperf2 branch.

PR #902, opened 2019-03-03
fixes #901 get/set function naming consistency

### LINKED ISSUE #901: Stream get/set functions naming consistency
## NNG & Platform details.
HEAD

## Expected/Actual Behavior
The `nng_stream` family of get/set methods have a few differences from the rest of the API:
- should be `getopt`/`setopt`
- `_addr` should be `_sockaddr` (this is __not__ fixed in the PR)

It's a little annoying when automagically generating things to have to deal with special cases (e.g. `get`/`opt` instead of `getopt`/`setopt` like everywhere else).

jake-ruyi:
fixes #901 Stream get/set functions should be getopt/setopt

Just a global search and replace.  Renamed adoc files as well.

### DIFF
--- a/src/core/stream.c
+++ b/src/core/stream.c
@@ -282,86 +282,10 @@ nni_stream_checkopt(const char *scheme, const char *name, const void *data,
 	return (NNG_ENOTSUP);
 }
 
-//
-// This next block sets up to define the various typed option functions.
-// To make it easier to cover them all at once, we use macros.
-//
+NNI_DEFGETALL_PTR(stream)
+NNI_DEFGETALL_PTR(stream_dialer)
+NNI_DEFGETALL_PTR(stream_listener)
 
-#define DEFGET(base)                                                         \
-	int nng_##base##_get(                                                \
-	    nng_##base *s, const char *nm, void *vp, size_t *szp)            \
-	{                                                                    \
-		return (nni_##base##_getx(s, nm, vp, szp, NNI_TYPE_OPAQUE)); \
-	}
-
-#define DEFTYPEDGET(base, suffix, type, nnitype)                     \
-	int nng_##base##_get_##suffix(                               \
-	    nng_##base *s, const char *nm, type *vp)                 \
-	{                                                            \
-		size_t sz = sizeof(*vp);                             \
-		return (nni_##base##_getx(s, nm, vp, &sz, nnitype)); \
-	}
-
-#define DEFGETALL(base)                                        \
-	DEFGET(base)                                           \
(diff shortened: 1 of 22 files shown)


src/core/stream.c
```diff
@@ -317,34 +317,34 @@ DEFGETALL(stream_dialer)
...
+	int nng_##base##_setopt_string(                                  \
 	    nng_##base *s, const char *nm, const char *v)             \
 	{                                                             \
 		return (nni_##base##_setx(s, nm, v,                   \
 		    v != NULL ? strlen(v) + 1 : 0, NNI_TYPE_STRING)); \
 	}
 
 #define DEFSOCKADDRSET(base)                                      \
-	int nng_##base##_set_adddr(                               \
+	int nng_##base##_setopt_addr(                               \
```
jake-ruyi: Not sure how I noticed this.  Function doesn't seem to be used anywhere (yet).

gdamore:
So I made a conscious decision to use shorter names this go around.  I agree that the discrepancy is unfortunate.  I plan to go back and add the new style names to the legacy APIs, and then to deprecate the legacy ones.

Shorter names make the consuming code a little cleaner in my opinion, and while 3 letters isn't a lot, if it causes an 80-column wrap, its annoying.

jake-ruyi:
For docs, you want both sets of functions listed, just the new/shorter version, or something else?

gdamore:
I'll look at this later... but a few thoughts so far.

1. It needs corresponding doc changes.  The old APIs should be marked deprecated somehow in the docs, with references to the new ones.

2. The old APIs should be implemented as wrappers around the new ones, possibly using a similar macro strategy.  The reason to do this is to maximize test coverage, given that the old tests all use the old APIs.

jake-ruyi:
1. AsciiDoc doesn't seem to have a standard way to mark something as deprecated.  Can I get you to do one as an example and then I'll do the rest the same.
1. I didn't bother with this because both the old and new APIs are just wrappers around `nni_*_getx` and `nni_*_setx`.  If it's ok to re-implement them might as well macro-ify them as well.

jake-ruyi:
@gdamore How about doing the docs like this?
List both sets of functions, with a simple `DEPRECATED` comment about the `getopt`/`setopt` variants.

gdamore:
I probably would just have a NOTE, or maybe create a second man page, with the old one getting a NOTE that the functions are deprecated with a reference pointing to the new ones.

jake-ruyi:
Ok, created new man pages with `NOTE` leading from getopt/setopt

jake-ruyi:
@gdamore any more thoughts on this?

PR #904, opened 2019-03-03
fixes #903 Stream get_string implementation missing

### LINKED ISSUE #903: Stream get_string implementation missing
## NNG & Platform details.
HEAD

## Expected Behavior
Undefined symbol `nng_stream.*_get_string`

jake-ruyi:
fixes #903 Stream get_string implementation missing

This will probably conflict with #901.  In case the get/set naming is intentional and desired, making this independent of that.
If #901 is accepted, just make the one line change from here and delete this PR.  =)


gdamore:
I'm going to reject #901 -- watch there.  This one looks fine though -- it was an oversight, but strings are kind of special in the handling, because they are passed in by value, but strduped on the way out.  But examination reveals to me at least that this looks good.

Issue #862, opened 2019-01-18
Pair v1 Polyamorous misdirected messages


jake-ruyi:
## NNG & Platform details.

HEAD.  macOS.

## Expected Behavior

Based off [docs](https://nanomsg.github.io/nng/man/v1.1.0/nng_pair.7) and [tests](https://github.com/nanomsg/nng/blob/d6bb25e1e0a25cb5aa781ac4f90b513fd5624f50/tests/pair1.c#L196), if:
- `0` is a pair v1 with POLY and listening
- `1a` and `1b` are separate aio using another pair v1 and dialing
Then:
1. `1a` sends nng_msg `m1a0` to `0`
1. `0` receives `m1a0` and calls `nng_msg_dup` on it to make `m01a`
1. `0` calls `nng_msg_get_pipe(m1a0)` and `nng_msg_set_pipe(m01a)` and sends it
1. `1b` simultaneously does the same with `m1b0` (resulting in `m01b`)
1. `1a` receives `m01a` and `1b` receives `m01b`

Basically, calling `nng_msg_set_pipe` and `nng_send_aio` will deliver it back to where it came from.

## Actual Behavior

Sometimes they switch.  Meaning `1a` receives `m01b` and `1b` receives `m01a`.

jake-ruyi:
I originally ran into this with Rust, but my (dubious) C example is similar.
[main.c.txt](https://github.com/nanomsg/nng/files/2772393/main.c.txt)

gdamore:
I suspect that the problem here may be handling of duplicated messages. I'll see if I can figure this out.

jeikabu:
Thought I'd give that a try since it's an easy fix.  Switched `nng_msg_dup` for allocating a new message and it does the same.  I'll try it back in my original test in case there's a problem with the C repro I put together.
Updated: 
[main.c.no_dup.txt](https://github.com/nanomsg/nng/files/2777492/main.c.no_dup.txt)

gdamore:
You wouldn't happen to have a C based reproduction of this would you?

jeikabu:
> [main.c.no_dup.txt](https://github.com/nanomsg/nng/files/2777492/main.c.no_dup.txt)

It's attached

jake-ruyi:
> [main.c.no_dup.txt](https://github.com/nanomsg/nng/files/2777492/main.c.no_dup.txt)

It's attached.

jake-ruyi:
Here's a simpler version:
```c
[...]

PR #909, opened 2019-03-12
nng_msg_dup correctly duplicates pipe (mentioned in #862)

jake-ruyi:
This was mentioned as a possible cause of #862, but doesn't actually fix it.
A couple dozen of messages can be exchanged without issue, but eventually one gets delivered through the wrong pipe.


gdamore:
I had planned to make this change myself... the only reason I've not merged yet is I want to test to see if it actually addresses the polyamorous misdelivery.

gdamore:
Your test case for #862 is busted,  but we do need to copy the pipe as part of message duplication.

PR #920, opened 2019-04-10
fixes #919 Polling on subscriber socket recvfd seems broken

### LINKED ISSUE #919: Polling on subscriber socket recvfd seems broken
our contribution guidelines.

## NNG & Platform details.

master branch commit c3e062661388f70386d6766e3ce648030af340ee

```bash
$ uname -a
Linux parallels-Parallels-Virtual-Platform 4.15.0-47-generic #50~16.04.1-Ubuntu SMP Fri Mar 15 16:06:21 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

## Expected Behavior
Calling poll() with `pfd.events = POLLIN` using the file descriptor given via `NNG_OPT_RECVFD` on a sub0 socket should return with `(pfd.revents & POLLIN) != 0` after data is written on the connected publisher.
## Actual Behavior
The `poll()` call will not return or will return after a timeout period if one is given.

## Steps to Reproduce
See here: https://github.com/nanomsg/nng/pull/920

bsirang:
`sub0_recv_cb()` was not calling `nni_pollable_raise()` on `sock->recvable`.


bsirang:
@gdamore Please review. I don't understand the internals of nng well enough to know if the `nni_pollable_raise()` call needs to be placed elsewhere, but this seems to have solved the issue I encountered.

src/protocol/pubsub0/sub.c
```diff
@@ -414,7 +414,7 @@ sub0_recv_cb(void *arg)
 	if (msg != NULL) {
 		nni_msg_free(msg);
 	}
-
+	nni_pollable_raise(sock->recvable);
```
gdamore: This isn't quite right, but it's close.  The issue here is that we don't want to raise this FD unless we passed the SUB filter.  So what needs to happen is up around line 382 we need to set a flag about whether this was matched or not, and then only do the raise down here if we matched.  Otherwise we may wind up doing a lot more spurious wakeups.

bsirang:
@gdamore Please review again.

Note that I added the flag below this check as well
```
		if (ctx == nni_list_last(&sock->ctxs)) {
			dup = msg;
			msg = NULL;
		} else if (nni_msg_dup(&dup, msg) != 0) {
			continue; // TODO: Bump a stat!
		}
```

Because if `nni_msg_dup()` fails, it seemed like we also wouldn't want to raise the pollable fd. I updated the unit test to test both cases (subscriber topic matches vs. doesn't match)

PR #932, opened 2019-04-23
fixes #931 nng_ctx_send can block forever

### LINKED ISSUE #931: Call to `nng_ctx_send` blocks forever.
our contribution guidelines.

## NNG & Platform details.

* NNG v1.1.1 and master branch (as of 98655cdaef0f2d1434e12f21793c461a12dbe329)
* Linux ... 4.15.0-42-generic #45~16.04.1-Ubuntu SMP Mon Nov 19 13:02:27 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
* Rust, calling into the C library.

## Expected Behavior

* `nng_ctx_send` should return at some point. Ideally immediately.

## Actual Behavior

* The call to `nng_ctx_send` blocks indefinitely.

I am running a single REQ socket with five AIO/Context pairs. There is about fifteen seconds between sending a request and receiving a reply and each context sends a new request as soon as it is available. In other words, I am sending out approximately 20 requests per minute. After running this way for around two hours, the send function blocks.

## Steps to Reproduce

It is a pretty convoluted system so I may or may not be able to come up with a minimal test program. However, I do have the following:

* [The list of threads](https://github.com/nanomsg/nng/files/3105038/threads.txt)
* [Backtraces for threads that aren't waiting on condition variables](https://github.com/nanomsg/nng/files/3105037/thread-bt.txt)

The thread that seems to be the source of the problem is thread 34. It is a thread that I spawned and it is currently blocked in a call to `nng_ctx_send`. Specifically, the thread enters the loop at `idhash.c:106` and it never finds an exit condition. I tested this by setting a breakpoint at the start of the loop, a breakpoint at both return statements, and then ignoring the first breakpoint for a bit. The breakpoint at the start is hit 277074 times and neither of the breakpoints at the return statements are ever hit.

The loop repeats through the following conditions:

```
id = 3683819696
h->ih_entries[0] = {ihe_key = 18446744073709551615, ihe_val = 0x0, ihe_skips = 1}
h->ih_entries[1] = {ihe_key = 18446744073709551615, ihe_val = 0x0, ihe_skips = 1}
h->ih_entries[6] = {ihe_key = 3683819690, ihe_val = 0x7fffd4000f80, ihe_skips = 1}
h->ih_entries[7] = {ihe_key = 3683819694, ihe_val = 0x7fffd40037b0, ihe_skips = 1}
h->ih_entries[4] = {ihe_key = 18446744073709551615, ihe_val = 0x0, ihe_skips = 1}
h->ih_entries[5] = {ihe_key = 3683819695, ihe_val = 0x7fffd4002d10, ihe_skips = 1}
h->ih_entries[2] = {ihe_key = 3683819685, ihe_val = 0x7fffd4000a60, ihe_skips = 1}
h->ih_entries[3] = {ihe_key = 18446744073709551615, ihe_val = 0x0, ihe_skips = 1}
```

neachdainn:
fixes #931 `nng_ctx_send` can block forever

This implements the first solution I suggested in the issue (having the `for` loop exit after probing all slots). I have not had a chance to test this in the system that I originally found the bug in but I will do so tomorrow.

### DIFF
--- a/src/core/idhash.c
+++ b/src/core/idhash.c
@@ -98,11 +98,13 @@ static size_t
 nni_hash_find_index(nni_idhash *h, uint64_t id)
 {
 	size_t index;
+	size_t start;
 	if (h->ih_count == 0) {
 		return ((size_t) -1);
 	}
 
 	index = NNI_IDHASH_INDEX(h, id);
+	start = index;
 	for (;;) {
 		// The value of ihe_key is only valid if ihe_val is not NULL.
 		if ((h->ih_entries[index].ihe_key == id) &&
@@ -113,7 +115,13 @@ nni_hash_find_index(nni_idhash *h, uint64_t id)
 			return ((size_t) -1);
 		}
 		index = NNI_IDHASH_NEXTPROBE(h, index);
+
+		if (index == start) {
+			break;
+		}
 	}
+
+	return ((size_t) -1);
 }
 
 static int


neachdainn:
I can confirm that this fixes the issue for me.

gdamore:
I like this change, but I don't think it's complete.

In particular, I don't think we are growing when we should.  I will probably integrate this change, and follow up with another change to ensure that we do grow when we need to.

PR #939, opened 2019-05-11
Add uint64 to generated legacy nng_setopt_* functions.

codypiersall:
The functions nng_dialer_setopt_uint64 and nng_dialer_setopt_uint64 are
declared in nng.h but not defined, causing errors at runtime of programs
that expect them to be defined.


codypiersall:
This was causing [pynng](https://github.com/codypiersall/pynng) to be unable to work against master.

PR #943, opened 2019-05-17
ZeroTier transport changes

janjaapbos:
Fix for:
- zt_ep_set_add_local_addr
- zt_ep_set_clear_local_addrs

New methods to retrieve udp address used for ZeroTier:
- zt_ep_get_udp4_addr
- zt_ep_get_udp6_addr

contributing guidelines.

### DIFF
--- a/src/transport/zerotier/zerotier.c
+++ b/src/transport/zerotier/zerotier.c
@@ -2958,6 +2958,7 @@ zt_ep_get_udp4_addr(void *arg, void *data, size_t *szp, nni_type t)
 
 	struct sockaddr_in *ssin = (struct sockaddr_in *)&ss;
 
+	sin.s_in.sa_addr = ssin->sin_addr.s_addr;
 	sin.s_in.sa_port = ssin->sin_port;
 
 	return (nni_copyout_sockaddr(&sin, data, szp, t));
@@ -2982,7 +2983,8 @@ zt_ep_get_udp6_addr(void *arg, void *data, size_t *szp, nni_type t)
 
 	struct sockaddr_in6 *ssin = (struct sockaddr_in6 *)&ss;
 
-	sin.s_in.sa_port = ssin->sin6_port;
+	memcpy(&sin.s_in6.sa_addr, ssin->sin6_addr.s6_addr, 16);
+	sin.s_in6.sa_port = ssin->sin6_port;
 
 	return (nni_copyout_sockaddr(&sin, data, szp, t));
 }
--- a/src/transport/zerotier/zerotier.c
+++ b/src/transport/zerotier/zerotier.c
@@ -14,6 +14,10 @@
 #include <string.h>
 
 #include "core/nng_impl.h"
+#include "core/platform.h"
+
+#include "platform/posix/posix_pollq.h"
+#include "platform/posix/posix_udp.h"
 
 #include "nng/transport/zerotier/zerotier.h"
 
@@ -56,6 +60,7 @@ typedef struct zt_ep       zt_ep;
 typedef struct zt_node     zt_node;
 typedef struct zt_frag     zt_frag;
 typedef struct zt_fraglist zt_fraglist;
+
 
 // Port numbers are stored as 24-bit values in network byte order.
 #define ZT_GET24(ptr, v)                              \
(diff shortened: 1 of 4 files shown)


src/transport/zerotier/zerotier.c
```diff
@@ -2706,13 +2710,13 @@ zt_ep_set_add_local_addr(void *arg, const void *data, size_t sz, nni_type t)
 		case NNG_AF_INET:
 			sin                  = (void *) &ss;
 			sin->sin_family      = AF_INET;
+			sin->sin_port        = sa.s_in.sa_port;
```
gdamore: My decision *not* to fix the local port was intentional.  Do you have a need for this?

janjaapbos: Yes, this one I need. The local addresses that are communicated to peers must have both the ip and port (so the peers can try to directly contact at that ip/port). This is why I also made a method to retrieve the random selected udp port that is created for the zt udp transport.

janjaapbos: By informing peers of the local addresses to try, there is a higher success of direct connections instead of being relayed by the zerotier root servers, especially if the peers are behind the same nat firewall.

gdamore: Ok, I may have totally misunderstood what we are trying to do.

The "local address" I thought to mean the local address on the system -- meaning our *own* IP address.  The value zero was meant to mean that the system should attempt to bind to any available port, not fix to a single hard coded port.

If the local address in ZeroTier means something different, like the address of a peer on the same subnet, that's a completely different thing, and the original code was bad.

It would be good to get full clarification here.

janjaapbos: You are talking about the wildcard listen address of the udp transport, which is set up in zt_node_create. 

The zt_ep_set_add_local_addr function only informs the zerotier node / transport of the different ip addresses that peers can use to reach the node. (I have in my broker program a function that enumerates all the ip address active on the host, and calls zt_ep_set_add_local_addr).

janjaapbos: So zt_ep_set_add_local_addr is called with the different IPs on the host, with the port used for the wildcard listen.

janjaapbos: I retrieve the port with zt_ep_get_udp4_addr before I call zt_ep_set_add_local_addr.

janjaapbos: See for example: https://github.com/capitar/capitar-zerotier-broker/blob/393e2d01d4c441bbd3036e05b919614238ffdeb5/util.c#L321

gdamore: Ok, that helps clarify it for me.  I do wonder if it would be better to use a single option to get both the v4 and v6 addresses -- we pretty much don't make any such distinction anywhere else in the API.

janjaapbos: Yes, that make sense. I'll update the pull request with that.

janjaapbos: This more complicated than I thought. It is not possible to have both an ipv6 and and an ipv4 address / port in nng_sockaddr, as it is a union of these.

So I will have to seek another data type to return, as I cannot use nng_sockaddr for this.

Perhaps keep it just as it is?

janjaapbos: For completeness I changed it to also store the ip address, not only the port.

gdamore: I may change this somewhat... it's clear that what I was confused about this when I did the original work.

My intention for address retrieval was that this would return an array (just two) of nng_sockaddrs, not overload a single one.

janjaapbos: Sounds good. Thanks!

src/transport/zerotier/zerotier.c
```diff
@@ -14,6 +14,10 @@
 #include <string.h>
 
 #include "core/nng_impl.h"
+#include "core/platform.h"
+
+#include "platform/posix/posix_pollq.h"
+#include "platform/posix/posix_udp.h"
```
gdamore: The above headers should not be included in platform independent code.

gdamore:
I know things have changed a lot.

I still plan to revisit ZT in general.  I'm convinced that 6PLANE is the way to go in the long run.

However, in the meantime I need to look at more modest changes like this.

Thanks for your contribution -- I will be following up.

janjaapbos:
I will do rewrite of these methods for setting and getting udp address, taking into account the comments above.

dch:
FWIW I use ZT 6PLANE to link up our containers across servers and different providers. It's so so sweet to use. @janjaapbos it would be great to see some more ZT+NNG code examples about the place to look through, if you have some to share.

PR #949, opened 2019-05-19
fixes #948 TLS listener leaks underlying TCP stream listener

### LINKED ISSUE #948: TLS listener leaks underlying TCP stream listener
Found via valgrind.

gdamore:


gdamore:
Note that this leak was noticed with both the TLS and wss tests.

PR #950, opened 2019-05-19
Simply reap logic slightly.

gdamore:
The separate reap_empty variable brought no value -- a simpler
check of the actual list is easier to prove for correctness.

contributing guidelines.


gdamore:
So further testing... intersting behaviors noticed -- using the linked list being empty lead to occasional timeouts, or even crashes.  We should probably reexamine what we are using the reap_drain() for a bit more closely.

gdamore:
This is too busted to continue.  Closing it.

PR #952, opened 2019-05-23
fix compile error

trevor211:
fix compile error


gdamore:
Sorry, I should have merged this PR, but I fixed it in parallel thanks to someone else pointing it out, so the change is in.

PR #971, opened 2019-07-18
fixes #954 large message fails with no error message

### LINKED ISSUE #954: large message fails with no error message
our contribution guidelines.

## NNG & Platform details.

version: v1.1.1
platform centos 7
language: c

## Expected Behavior

nng_sendmsg should report an error when a message fails to send because it is too big, the return value should be non 0.
or
nng_recvmsg should report an error when a message cant be received because it is too big, the return value should be non zero.

## Actual Behavior
No error reported. No way to track down the issue.

## Steps to Reproduce

- Create a pair connection.
- Between two processes.
- Send a large message from one end to the other.
- The message fails to deliver silently.
- Now set nng_setopt_size(socket, NNG_OPT_RECVMAXSZ, 0);
- The message delivers with no problem.

gdamore:
This removes the default 1MB limit on maximum receive sizes.
Applications intended for deployment in insecure or hostile
environments should choose a sensible default for NNG_OPT_RECVMAXSZ.

### DIFF
--- a/docs/man/nng_options.5.adoc
+++ b/docs/man/nng_options.5.adoc
@@ -190,6 +190,9 @@ large message, without sending any data.
 This option can be set for the socket, but may be overridden for on a
 per-dialer or per-listener basis.
 +
+IMPORTANT: Applications on hostile networks should set this to a non-zero
+value to prevent denial-of-service attacks.
++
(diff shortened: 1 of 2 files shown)


gdamore:
Not happy about the test failure in IPC on Windows, but it isn't related to this change.

PR #973, opened 2019-07-29
Fix/prototypes

gatzka:
This is a small cleanup PR.
Some functions that are used only locally where not declared static. By making them static the compiler
can easier inline them. In addition, the compiler now checks if those functions are really used.


src/core/panic.c
```diff
@@ -20,7 +20,7 @@
 #include "core/nng_impl.h"
 
 // Panic handling.
-void
+static void
 nni_show_backtrace(void)
```
gdamore: This was external for a reason -- the nni_ prefix indicates symbols that are meant to be available for external linkage, but are not documented for use by other projects.  In particular I sometimes use this function during debugging.

gatzka: That's o.k. But then I would still expect a function src/core/panic.h
I can change this.

gatzka:
> I'm not entirely sure what you're trying to achieve here. Is this just to enable addition of a new warning?

Well, at the end, it is, yes. I find this check particularly useful because it points out non-static functions without prototypes. If the the prototype was just forgotten, well, it should be added to a suitable header file. If the function is not to be used outside the compilation unit, it should be made static to:
- not polluting the namespace
- allow inlining by the compiler
- to point out unused functions

So if you agree, I would change the pull request to put void
`void nni_show_backtrace(void)` into src/core/panic.h and resubmit.

gdamore:
So, I am going to close this for now.  I would consider a similar change that restricted itself to just t he library.  I have no interest in worrying about the scope of symbols in the test suites, or utilities -- they are not going to affect user programs.  I would be far more interested in scope management around the library itself.

Additionally, the change needs to be resubmitted without conflicts.

PR #974, opened 2019-08-06
Set a rcvmax for ZeroTier transport if 0

janjaapbos:
ZeroTier transport cannot deal with an unlimited rcvmax.

contributing guidelines.

### DIFF
--- a/src/transport/zerotier/zerotier.c
+++ b/src/transport/zerotier/zerotier.c
@@ -1689,7 +1689,8 @@ zt_pipe_alloc(
 	p->zp_proto      = ep->ze_proto;
 	p->zp_nwid       = ep->ze_nwid;
 	p->zp_mtu        = ep->ze_mtu;
-	p->zp_rcvmax     = ep->ze_rcvmax;
+	/* We need to set a rcvmax for ZeroTier transport, cannot be 0 (unlimited) */
+	p->zp_rcvmax     = ep->ze_rcvmax ? ep->ze_rcvmax : 1024 * 1024 * 1024;
 	p->zp_ping_tries = ep->ze_ping_tries;
 	p->zp_ping_time  = ep->ze_ping_time;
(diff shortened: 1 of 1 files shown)


src/transport/zerotier/zerotier.c
```diff
@@ -1689,7 +1689,8 @@ zt_pipe_alloc(
 	p->zp_proto      = ep->ze_proto;
 	p->zp_nwid       = ep->ze_nwid;
 	p->zp_mtu        = ep->ze_mtu;
-	p->zp_rcvmax     = ep->ze_rcvmax;
+	/* We need to set a rcvmax for ZeroTier transport, cannot be 0 (unlimited) */
+	p->zp_rcvmax     = ep->ze_rcvmax ? ep->ze_rcvmax : 1024 * 1024 * 1024;
```
gdamore: I want to review this -- a GB is probably wrong here.  Stay tuned.

gdamore: There are a couple of errors where the code that handles zp_recvmax incorrectly assumes a positive value. I need to fix that.  In theory 1GB is too small, but it's also really really too big for a successful result IMO -- we don't have good streaming here, and having a 1GB window is horrible.

Ultimately, I am coming to beleive we need to switch this out for something better, perhaps a KCP style thing on top of NNG, but that would be very much a breaking change.

I'll probably just accept this change for now, but I'm going to leave open a request to revise this and perhaps adopt a different approach altogether.

gdamore:
Actually, this is wrong for other reasons.  See #978.  The protocol is going to need to drop or error any message that absolutely is too large, but beyond that we should permit a maximum number of fragments to be supplied.  Stay tuned for follow up; in the meantime I'm closing this.

src/transport/zerotier/zerotier.c
gdamore: NVM.  I took a different approach … see #980

PR #980, opened 2019-08-10
fixes #979 ZeroTier fails with zero recvmax

### LINKED ISSUE #979: ZeroTier fails with zero recvmax
A bunch of code fails is the maximum receive is unset, which is the new default in the core.

See #974 for one attempt to address this.  A better solution (for now) is to properly check for zero when checking for limits.    We also need to recognize a default maximum number of fragments of 65535.

gdamore:
This should be a better solution than #974 --  testing still needed.


janjaapbos:
Thank you! I have tested these change with success.

PR #981, opened 2019-08-10
fixes #970 TCP connections through docker-proxy do not reconnect

### LINKED ISSUE #970: TCP connections through docker-proxy do not reconnect
## NNG & Platform details.
NG version: Current (master) branch (pulled July 11, 2019)
Platform:  Linux 4.15.0-51-generic #55-Ubuntu
Programming language:  C
CMake options used for NNG:  -DBUILD_SHARED_LIBS=1  

Other environmental factors: Processes are running in docker containers and communicating via docker-proxy.

## Expected Behavior
When a TCP connection between two processes is broken,  NNG reconnects when the failing process restarts.  (Processes are suing NNG-Push/Pull protcol running in docker containers, and networking is via docker-proxy.)

## Actual Behavior
NNG recognises the session failure, and attempts the reconnection **only a single time;** this happens long before the second process has recovered and restarted, and no additional attempts to connect are made.

Investigation shows that after the connection to the listener is broken,  NNG makes one connect() system call which is deferred (shows in progress)  and then succeeds establishing a 'half session' with the docker-proxy (SYN and a return of SYN-ack is observed).  We also observe that the pipe negotiation callback in transport/tcp/tcp.c is invoked multiple times,  and notices an error state (before the 10s timeout); error code in the function  is driven.  After this, the connect (dialer) code is never driven to reattempt the session. 

My theory was that because the half session appears active, NNG believes that the connection is good, but some state was not properly cleaned up when the error code in the pipe nego function was driven.   

Digging a bit further, I also observed that the dialer_timer callback (core/dialer.c) was only being driven once in this situation, and noticed that the NNG_ECLOSED/NNG_ECANCELED case  in the dialer_connect_cb() fuction was being taken; the comment there is no action is needed.  However, if I reset the d->d_started flag in the dialer, and schedule the dialer timer (as is done for the NNG_ETIMEDOUT case in the same switch, that NNG will retry the session and eventually establish a new connection when the second process restarts. 

I can submit a PR for this change, however I'm not sure that this is a complete fix, thus I'll start with a bug report.

## Steps to Reproduce
1) in two containers, started such that the exposed ports are available on the docker default network, start a 'listener' in one container, and a 'sender' in the other.

2) after sender establishes a TCP connection through the proxy to the listener, and messages are being exchanged, stop the listener process

3) restart the listener process, observe that the session does not resume. 

At present, I do not have a simple example, however I could build one of that becomes necessary.

gdamore:
This seems to pass the test suite, but I haven't yet tested this against the particular failure mode.

### DIFF
--- a/docs/man/nng_stream_recv.3str.adoc
+++ b/docs/man/nng_stream_recv.3str.adoc
@@ -56,6 +56,7 @@ None.
 `NNG_ECANCELED`:: The operation was canceled.
 `NNG_ECLOSED`:: The connection was closed.
 `NNG_ECONNRESET`:: The peer closed the connection.
+`NNG_ECONNSHUT`:: Remote peer shutdown after sending data.
 `NNG_EINVAL`:: The _aio_ does not contain a valid scatter/gather vector.
 `NNG_ENOMEM`:: Insufficient free memory to perform the operation.
 `NNG_ETIMEDOUT`:: Timeout waiting for data from the connection.
--- a/include/nng/nng.h
+++ b/include/nng/nng.h
@@ -233,7 +233,8 @@ NNG_DECL int nng_socket_set_uint64(nng_socket, const char *, uint64_t);
 NNG_DECL int nng_socket_set_string(nng_socket, const char *, const char *);
 NNG_DECL int nng_socket_set_ptr(nng_socket, const char *, void *);
 NNG_DECL int nng_socket_set_ms(nng_socket, const char *, nng_duration);
-NNG_DECL int nng_socket_set_addr(nng_socket, const char *, const nng_sockaddr *);
+NNG_DECL int nng_socket_set_addr(
+    nng_socket, const char *, const nng_sockaddr *);
 
 NNG_DECL int nng_socket_get(nng_socket, const char *, void *, size_t *);
 NNG_DECL int nng_socket_get_bool(nng_socket, const char *, bool *);
@@ -349,7 +350,8 @@ NNG_DECL int nng_dialer_set_uint64(nng_dialer, const char *, uint64_t);
 NNG_DECL int nng_dialer_set_string(nng_dialer, const char *, const char *);
 NNG_DECL int nng_dialer_set_ptr(nng_dialer, const char *, void *);
 NNG_DECL int nng_dialer_set_ms(nng_dialer, const char *, nng_duration);
-NNG_DECL int nng_dialer_set_addr(nng_dialer, const char *, const nng_sockaddr *);
+NNG_DECL int nng_dialer_set_addr(
+    nng_dialer, const char *, const nng_sockaddr *);
 
 NNG_DECL int nng_dialer_get(nng_dialer, const char *, void *, size_t *);
 NNG_DECL int nng_dialer_get_bool(nng_dialer, const char *, bool *);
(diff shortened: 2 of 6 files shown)


ScottDaniels:
I've pulled and tested this branch.  It is partially working.

Case 1 - working:
Connection made through proxy to an application which accepts the connection, but does not successfully complete the handshake and terminates.

In this case the session disconnect is noticed on the FIN, and I see SYNs start to go out until the process on the other side of the proxy returns. If the other process successfully completes the handshake the session is begins and messages are exchanged as expected.

Case 2 - not working:
Connection is made, handshakes are exchanged, and data begins to flow.  When the "receiver" is terminated, I see that NNG on the sender side recognise the FIN, and immediately send a SYN packet, but unlike the working case I never see another SYN flow, so when the receiver is started a session is never reestablished.

The expectation was that periodic SYNs would be issued as was happening in the case where the NNG setup had not completed successfully.

I will continue to investigate with the hope of providing additional details, but thought this small bit might be useful.

gdamore:
Interesting.  I suspect that there is somewhere where NNG_ECLOSED handling happens, but where NNG_ECONNSHUT also needs to be handled.

ScottDaniels:
I've verified that for case 2 (above)  the test in src/platform/posix/posix_tcpconn.c @179 is seeing 0 bytes and cancelling the aio with NNG_ECONNSHUT.  Just seems that a retry isn't being scheduled.

ScottDaniels:
Not sure if this helps, but erroring on the side of adding useless info than keeping it to myself since I think it's difficult for you to drive this particular situation.

I've continued to dig and what I see is immediately after the failrue, Negotiation begins, and ECONNSHUT is triggered. I see another call to tcp_doread() as I'd expect, except this results in an EAGAIN and the negotiation callback receives an ECLOSED state (7) which is causing the timer not to be reset.  

I still haven't been able to connect the dots between the EAGAIN observed by tcp_doread and what is calling tcp_close which is setting the state (or so it seems to me).

ScottDaniels:
A bit more digging and I think I've found the source of the ECLOSED state during the negotiation.   In tcp_cb() (src/platform/posix/posix_tcpconn.c @244) the test for some kind of poll error is true and the block with tcp_close(), which closes the aio with ECLOSED, is invoked. When I forced this one call to tcp_close() to generate a ECONNSHUT instead, the timeout is reset as expected, and the connection is established when the other process starts.

I'm assuming that if testing for a poll error, this isn't the case where the user application has closed the socket, and from my inexperienced point of view, it seems that if a poll error is treated in the same way as 0 bytes read, then the behavour is as desired. 

This works for the the case where both applications are on the same container network (no proxy) and when communicating through the docker proxy (original failrue case).  I'll test to ensure that if the application specifically closes the socket it isn't automatically reconnected.

ScottDaniels:
I've confirmed that setting the NNG_ECONNSHUT state when tcp_cb() detects a poll error does **not** affect an application initiated close via nng_close().   Tested both with and without docker proxy.   

I think adding this completes the change to fix the bug.    Happy to test further if/when needed.

gdamore:
@ScottDaniels I've just committed a change that I think should resolve the other case.  If you have a chance to test, I'd be grateful.

gdamore:
(Still on this PR though.)

ScottDaniels:
Fantastic!   I should have a chance late today or in the morning tomorrow.   (US-EDT).

ScottDaniels:
**Confirmed!**  Pull and build of this PR now **fixes** the issue.     I think you're good to go with the merge.

PR #997, opened 2019-09-25
Add option to SUB0 protocol to prefer newer messages when queue is full

neachdainn:
This adds a new option (`NNG_OPT_SUB_PREFNEW`) that toggles whether SUB0 contexts ignore new messages when the queue is full or whether they drop an old message in order to make room.

This probably needs some work to be ready to merge into master, so I'm more than willing to make whatever changes necessary.

### DIFF
--- a/docs/man/nng_sub.7.adoc
+++ b/docs/man/nng_sub.7.adoc
@@ -73,6 +73,13 @@ TIP: To receive all messages, an empty topic (zero length) can be used.
    Note that if the topic was not previously subscribed to with
    `NNG_OPT_SUB_SUBSCRIBE` then an `NNG_ENOENT` error will result.
 
+((`NNG_OPT_SUB_PREFNEW`))::
+
+   (`bool`)
+   This read/write option specifies the behavior of the subscriber when the queue is full.
+   When `true` (the default), the subscriber will make room in the queue by removing the oldest message.
+   When `false`, the subscriber will reject messages if the message queue does not have room.
+
 === Protocol Headers
 
 The _sub_ protocol has no protocol-specific headers.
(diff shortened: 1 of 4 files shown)


neachdainn:
I plan on adding a similar option to the PUB0 protocol but I assumed it would be better to tackle them independently. However, I don't know if the option key should be shared or whether having both a `NNG_OPT_SUB_PREFNEW` and a `NNG_OP_PUB_PREFNEW` is fine.

gdamore:
This all looks quite good, but I think you should update the copyright headers appropriately, and squash the commits into one, and then I will be happy to approve.

neachdainn:
Just to be clear, "update the copyright headers" means including a line with me on it, right? And having both `NNG_OPT_{PUB,SUB}_PREFNEW` is your preferred way of handling a similar change in PUB0?

gdamore:
Yes on the copyright.  I am not sure that there is any reason to do this on the pub side as there is no queue there really.

neachdainn:
I was confused about that when I started looking into this. Aren't there message queues in the publisher's pipes that can potentially fill? And based on [this line][1] it looks like they'll start rejecting newer messages if full. Is there some mechanism I'm not seeing that either prevents this or otherwise makes a change similar to this PR undesirable?

[1]: https://github.com/nanomsg/nng/blob/master/src/protocol/pubsub0/pub.c#L270

gdamore:
Good point.  To be honest, the queue there is deeper than I remember.

In retrospect, I think the default for this option should be true -- meaning we should always prefer the new messages.  Old messages sitting around are likely to be stale and largely useless.

Can we change this PR to do that please?

neachdainn:
> Can we change this PR to do that please?

Consider it done! I only made `false` the default as that was the legacy behavior.

> Good point. To be honest, the queue there is deeper than I remember.

If we're swapping the default on the sub side, do you want me to just implement this same behavior in the pub side as part of this PR?

gdamore:
Yes, that would be fine.

neachdainn:
Ok. I'll probably get to that tonight but it might not be for an hour or so.

neachdainn:
That latest push includes the changes to the pub socket as well.

PR #1015, opened 2019-11-10
Add support for clang-format custom target

hugolm84:
This PR allows a user to format all source files via the custom target `format`. It uses the `.clang-format` file already supported by the project. 
Enable by passing in `-DCLANG_FORMAT=ON` when configuring CMake, specify `format` as build target: `cmake --build . --target format`.

contributing guidelines.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -394,6 +394,24 @@ endif ()
 
 add_subdirectory (docs/man)
 
+find_package(ClangFormat 9.0 EXACT)
+if(ClangFormat_FOUND)
+    set(search_dirs demo etc include perf src test tools)
+    foreach(dir IN LISTS search_dirs)
+        list(APPEND search_paths "${CMAKE_SOURCE_DIR}/${dir}/*.c")
+        list(APPEND search_paths "${CMAKE_SOURCE_DIR}/${dir}/*.h")
+    endforeach()
+    file(GLOB_RECURSE FORMAT_SOURCE_FILES ${search_paths})
+    add_custom_target(format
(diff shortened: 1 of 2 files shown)


gdamore:
I actually just use an editor plugin for this.  I'm not sure how I feel about carrying this baggage around, because I don't think anyone will really use it.

What might be constructive is figuring out how to run clang-format on sources prior to PR as one of the PR checks.

hugolm84:
Hi

Today, formatting is a manual labour that you and anyone else that want to contribute needs to do using a supported editor.
Making it a automated step would make I easier for anyone, be it by using it as a pre commit hook, PR check, running it before or after build in the terminal or using the editor (should you use one).
The point is, code consitency is important and instead of discussing style during review, you could direct all efforts to the changes in code and this PR helps in that direction.

I did not want to make changes to how it is enforced, but felt giving the project the tools needed, usage could be discussed.

The only thing that I would consider bagage here is maintaining the list of directories for sources.
Enforcing a certain version of clang-format is important so that every contributor produces the same format changes.

Running format on this codebase, formats a few unformatted files, which proves my point and therefore makes me believe this PR is useful.

gdamore:
I'm going to decline this PR for now.  I do think we should revisit the idea generally though -- but I'm not thrilled with further complicating our CMake stuff -- its got plenty of baggage as it is.

What instead I'd like to see about is a git hook or possibly a check in the form of a GitHub action, that did this. 

At the same time, most developers have ready access to editors with clang-format, or can run clang-format on the command line.  I believe the days where this was an onerous requirement for contribution are behind us.

PR #1028, opened 2019-12-16
add demo/reqrep into automatic Cmake build

pedro-vicente:
This is an attempt to add demo/reqrep to the automatic Cmake build.
demo/reqrep/reqrep.c
was divided in 2 files/projects 
req.c
rep.c
to allow easy debugging with Visual Studio (running the 2 projects at the same time)


gdamore:
For some platforms, we can't build the demos, because e.g. there might not be a main().

gdamore:
So without conditionalizing the build of these, I can't accept it.  If you added a switch to disable these (we have such a switch for the tools for example) then I could reconsider.  Closing for now.

PR #1045, opened 2019-12-26
Add possibility to use mbedtls cmake target if available

robiwano:
If mbedTLS is added as a CMake target, this fix will handle it properly.

### DIFF
--- a/src/supplemental/tls/CMakeLists.txt
+++ b/src/supplemental/tls/CMakeLists.txt
@@ -29,9 +29,15 @@ if (NNG_SUPP_TLS_MBEDTLS)
         Consult a lawyer and the license files for details.
         ************************************************************")
 
-	find_package(mbedTLS REQUIRED)
-	set(_LIBS ${MBEDTLS_LIBRARIES})
-	set(_INCS ${MBEDTLS_INCLUDE_DIR})
+	# If mbedTLS was added by a consuming project, then we should use that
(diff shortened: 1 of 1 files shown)


gdamore:
How did you test that this actually gets  executed?  It seems on paper like this is probably a good approach, but I thought I was using the approach documented to work by the mbedTLS folks.

gdamore:
I guess I'd also like to better understand the motivation here -- is the use case in a larger project where you already have mbedTLS for other reasons?

robiwano:
I have written a small C++ REST server framework, wherein this is the way I enable TLS support. I use CMakes FetchContent mechanism to download the ARM mbedTLS package (when needed), and it is then added with add_subdirectory. So mbedtls ends up being a CMake target (hence the `if(TARGET mbedtls)`). I'm going to put it up soon on my github pages.

gdamore:
I feel like it would be helpful to have some comments explaining this inline with this -- CMakefiles tend to acquire a lot of cruft over time, and it can be hard to understand why some of this stuff is there later, which means it becomes at risk of removal.

Probably something like this:

# If mbedTLS was added by a consuming project's FetchContent, then we should use that
# instance of it, instead of configuring our own.

robiwano:
I've now put up my C++ framework here: https://github.com/robiwano/siesta/tree/feature/tls_support

PR #1047, opened 2019-12-27
Reference count TCP dialers. (WIP)

gdamore:
A similar change will be needed for IPC.  Need to investigate
IPC, and also TCP listeners and Windows.

contributing guidelines.


gdamore:
Running one more cycle, with IPC fixes, but if it passes I'm going to declare success. ;-)

PR #1049, opened 2019-12-27
fixes #940 httpclient Timeout can succeed in cloud

### LINKED ISSUE #940: httpclient Timeout can succeed in cloud
nng at 7eaccabfe

All tests are green on OSX, but on my Windows 2016 server vps, the httpclient test consistently fails on the "Timeout works" test. Line 38, Test: nng_aio_result(aio) == NNG_ETIMEOUT.

gdamore:

contributing guidelines.

### DIFF
--- a/tests/httpclient.c
+++ b/tests/httpclient.c
@@ -10,7 +10,6 @@
 
 // Basic HTTP client tests.
 
-
 #ifndef _WIN32
 #include <arpa/inet.h>
 #endif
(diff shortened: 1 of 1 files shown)


gdamore:
Basically this change attempts to tell any intermediate cache *not* to.  Because apparently example.com is cached and answered too darn quickly.

This is a hack until we provide our own testing server.

PR #1060, opened 2019-12-28
Add http server method for getting port number

robiwano:
When a http server is started with port zero, a port is automatically assigned. This PR just makes it possible to retrieve the port number.

### DIFF
--- a/docs/man/libnng.3.adoc
+++ b/docs/man/libnng.3.adoc
@@ -410,6 +410,7 @@ These functions are intended for use with HTTP server applications.
 |xref:nng_http_hijack.3http.adoc[nng_http_hijack()]|hijack HTTP server connection
 |xref:nng_http_server_add_handler.3http.adoc[nng_http_server_add_handler()]|add HTTP server handler
 |xref:nng_http_server_del_handler.3http.adoc[nng_http_server_del_handler()]|delete HTTP server handler
+|xref:nng_http_server_get_addr.3http.adoc[nng_http_server_get_addr()]|get HTTP server address
 |xref:nng_http_server_get_tls.3http.adoc[nng_http_server_get_tls()]|get HTTP server TLS configuration
 |xref:nng_http_server_hold.3http.adoc[nng_http_server_hold()]|get and hold HTTP server instance
 |xref:nng_http_server_release.3http.adoc[nng_http_server_release()]|release HTTP server instance
--- a/docs/man/nng_http_server_get_addr.3http.adoc
+++ b/docs/man/nng_http_server_get_addr.3http.adoc
@@ -1,0 +1,46 @@
+= nng_http_server_get_addr(3http)
+//
+// Copyright 2018 Staysail Systems, Inc. <info@staysail.tech>
+// Copyright 2018 Capitar IT Group BV <info@capitar.com>
+//
+// This document is supplied under the terms of the MIT License, a
+// copy of which should be located in the distribution where this
+// file was obtained (LICENSE.txt).  A copy of the license may also be
+// found online at https://opensource.org/licenses/MIT.
+//
+
+== NAME
+
+nng_http_server_get_addr - get HTTP server address
+
+== SYNOPSIS
+
+[source, c]
+----
+#include <nng/nng.h>
+#include <nng/supplemental/http/http.h>
+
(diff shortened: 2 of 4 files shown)


src/supplemental/http/http_public.c
```diff
@@ -779,6 +779,18 @@ nng_http_server_get_tls(nng_http_server *srv, struct nng_tls_config **cfgp)
 #endif
 }
 
+int
+nng_http_server_get_port(nng_http_server *srv, int *portp)
```
gdamore: The port numbers should be uint16_t.

It might be better to have a way to get the local address.  We have NNG_OPT_LOCADDR for other places in the code that use dynamically allocated ports like this.

robiwano: You mean something similar to nng_stream_listener_get_int?

gdamore: More like nng_stream_listener_get_addr().  This would give you the entire address (IP address and port) for the server.

The reason the server gets the port under the hood is to support websocket consumers that want to know the details.

Actually, Right now we can do this for websocket via the private nni_http_server_getx() API --
because websocket listeners (which are just HTTP servers) use this to support NNG_OPT_LOCADDR.

I'm about to commit test code which demonstrates this quite nicely.

robiwano: Ok, I'll wait for that and update the PR :)

gdamore:
This looks pretty good.

The only thing is that as this is introducing a new public symbol there are a few couple of things we should do:

1. Add your copyright statement to the changed files (this will be 2019 copyright)
2. Add documentation for it. 
3. Bump the minor version (this becomes v1.3 then)

I'm happy to deal with this in several ways:

1. On copyright, you can do this, or if you don't want to assert it, that's fine too.  If you're going to do it, please update this PR.

2.  On the docs, if you want to make the changes in this PR please do it.  If not, let me know, and I'll open a ticket to track the need.  Then either of us can follow up with that work later.

3. I'll probably go ahead and bump the version number once this change is merged.  (I'm on the cusp of release a 1.2.2 to accommodate other important fixes).

gdamore:
Just to be clear, please let me know how you want to proceed -- I'll wait for your reply before I merge this.

robiwano:
1. No need for my copyright.
2. As I'm unfamiliar with the doc system, I'd rather not do this :)

robiwano:
A thought, since other "get address functions" look like xxx_get_addr (instead of xxx_get_address) maybe the symbol should be nng_http_server_get_addr ?

robiwano:
Ah, ended up writing the docs. Asciidoc is quite nice, have mainly been using doxygen + markdown so far :)

docs/man/nng_http_server_get_addr.3http.adoc
```diff
@@ -0,0 +1,46 @@
...
+nng_http_server_get_addr - get HTTP server address
+
+== SYNOPSIS
+
+[source, c]
+----
+#include <nng/nng.h>
+#include <nng/supplemental/http/http.h>
+
+int nng_http_server_get_addr(nng_http_server *server, nng_sockaddr* sap);
```
gdamore: very minor spacing nit.  nng_sockaddr *sap  not nng_sockaddr* sap

docs/man/libnng.3.adoc
```diff
@@ -419,6 +419,7 @@ These functions are intended for use with HTTP server applications.
 |xref:nng_http_server_res_error.3http.adoc[nng_http_server_res_error()]|use HTTP server error page
 |xref:nng_http_server_start.3http.adoc[nng_http_server_start()]|start HTTP server
 |xref:nng_http_server_stop.3http.adoc[nng_http_server_stop()]|stop HTTP server
+|xref:nng_http_server_get_addr.3http.adoc[nng_http_server_get_addr()]|get HTTP server address
```
gdamore: This stanza was alphabetically sorted, so can you move this line up a few to it's correct alphabetical location?

gdamore:
Awesome -- thanks for updating with docs.  (And yes, asciidoctor is very nice to work with.)

Two super easy to fix nits in your docs.  Do you mind fixing them before I merge?

robiwano:
Fixed them now.

PR #1072, opened 2019-12-30
fixes #1071 tran_chkopt can be cleaned up

### LINKED ISSUE #1071: tran_chkopt can be cleaned up

gdamore:

contributing guidelines.


gdamore:
This is completely borked, and I don't want to think about it now.  I'll come back to it later.

PR #1077, opened 2019-12-30
Fix/nng enable stats

robiwano:
Fixes some compilation issues when NNG_ENABLE_STATS is set to OFF.

### DIFF
--- a/src/core/dialer.c
+++ b/src/core/dialer.c
@@ -128,6 +128,7 @@ dialer_stats_init(nni_dialer *d)
 void
 nni_dialer_bump_error(nni_dialer *d, int err)
 {
+#ifdef NNG_ENABLE_STATS
 	switch (err) {
 	case NNG_ECONNABORTED:
 	case NNG_ECONNRESET:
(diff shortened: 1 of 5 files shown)


robiwano:
Also adds nng_http_server_get_addr symbol to the CMakeLists.txt file for doc generation.

src/core/dialer.c
```diff
@@ -128,6 +128,9 @@ dialer_stats_init(nni_dialer *d)
 void
 nni_dialer_bump_error(nni_dialer *d, int err)
 {
+#ifndef NNG_ENABLE_STATS
+	NNI_ARG_UNUSED(d);
```
gdamore: Probably should have done this for err too.

src/core/listener.c
```diff
@@ -125,6 +125,9 @@ listener_stats_init(nni_listener *l)
 void
 nni_listener_bump_error(nni_listener *l, int err)
 {
+#ifndef NNG_ENABLE_STATS
```
gdamore: Also for err.

robiwano:
Hmm... one test is failing, not sure why. Investigating.

gdamore:
I don't know either.  The old Convey tests are hard to debug, and I've found them somewhat fragile in some of the cloud providers.  I'm in the process of converting them.  At any rate, I will waive the test failure, since it clearly isn't in the code you changed.

PR #1084, opened 2020-01-01
Add error message for NNG_ECONNSHUT

robiwano:
Adds error message for error NNG_ECONNSHUT.

### DIFF
--- a/src/nng.c
+++ b/src/nng.c
@@ -626,7 +626,8 @@ nng_setopt(nng_socket s, const char *name, const void *val, size_t sz)
 }
 
 static int
-nni_socket_getx(nng_socket s, const char *name, void *val, size_t *szp, nni_type t)
+nni_socket_getx(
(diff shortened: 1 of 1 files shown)


src/nng.c
```diff
@@ -626,7 +626,8 @@ nng_setopt(nng_socket s, const char *name, const void *val, size_t sz)
 }
 
 static int
-nni_socket_getx(nng_socket s, const char *name, void *val, size_t *szp, nni_type t)
+nni_socket_getx(
+    nng_socket s, const char *name, void *val, size_t *szp, nni_type t)
 {
```
robiwano: Interestingly, my clang-format has concluded this should be reformatted... (?)

gdamore: What version of clang-format are you using?  Sometimes new versions change things...

gdamore: Actually mine does too now.

PR #1085, opened 2020-01-02
fixes #1083 Random number improvements

### LINKED ISSUE #1083: Random number improvements
We're using ISAAC to provide CSPRNG.  However, ISAAC isn't approved for some uses, and our implementation requires locks and such to be available.  We can simplify things by using the platform provided random number generators when available.  Additionally, we have made some unfortunate choices in the past;

1. We should use arc4random in preference to all others when available.
2. We should use getrandom if arc4random is not available.
3. The getentropy() call is probably never useful.
4. rand_s() is to be used on Windows.

If we have no other choices, only *then* should we fall back to using /dev/urandom if present.  Arguably, using /dev/urandom is something we can *always* do if we're stuck in this situation, and not waste time with the whole ISAAC thing.

Given all this, it's kind of unclear to me whether we even need to keep the ISAAC code around.  It might be useful for folks porting to other platforms where they have to fabricate entropy from other sources (e.g. random hardware sources), but honestly any such platforms (likely embedded) probably have other RNG services that they can use more directly.

gdamore:
This cleans up the RNG, and makes use of the underlying operating system implementations, ditching our own ISAAC CSPRNG.  The platform code should be used as its potentially more secure, faster (it can avoid global locks such as by using TLS, or use spinlocks if needed) etc.

If your platfrom is crappy, you might get stuch with reading from /dev/urandom.  Ask your systems vendor to give you a modern arc4random implementation.  (HPUX, AIX, and IRIX looking at you.)

Even illumos and Solaris have arc4random these days.


gdamore:
Note that on POSIX-like systems without a working /dev/urandom (and none of the alternatives), the new code will panic during early startup.  We think that's better than lying about having good quality random numbers.

PR #1106, opened 2020-01-04
Aioclean WIP

gdamore:

contributing guidelines.


gdamore:
Declining this one -- I need to think carefully because there are some sticky things about tearing down the tasks.  In particular, objects that are going to be freed cannot be freed from a callback.  The streams code is guilty of destroying streams from connection callbacks, and in particular the TLS and negotiation handlers get this wrong.

We need to defer the finalizers in this case.

PR #1140, opened 2020-01-16
Fix URI test that fails on all non-base URIs

### LINKED ISSUE #751: http server directory handler

robiwano:
fixes #751 http server directory handler

The test `(uri[len] != '/')` fails on all URIs other than "/". This PR fixes this.

### DIFF
--- a/src/supplemental/http/http_server.c
+++ b/src/supplemental/http/http_server.c
@@ -676,7 +676,7 @@ http_sconn_cbdone(void *arg)
 	// Get the handler.  It may be set regardless of success or
 	// failure.  Clear it, and drop our reference, since we're
 	// done with the handler for now.
-	h   = nni_aio_get_data(aio, 1);
+	h = nni_aio_get_data(aio, 1);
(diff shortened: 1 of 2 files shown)


robiwano:
Ok, I see now. The problem is only with serving "/". Trying to figure out a solution now.

gdamore:
You've reopened, but there are test failures here.  I can possibly have a look, but if you're working on it I just assume let you finish it up.

robiwano:
Yes, closed it by mistake :) I'm giving it a go to see if I can make it work with "/".

robiwano:
Hmm... I saw something I need to fix.

robiwano:
Nah, never mind. The test should work fine as it passes through everything if base is "/"

gdamore:
Thanks for your contribution.

PR #1150, opened 2020-01-19
Add access-control-allow-origin header to http server response for ha…

robiwano:
…ndling CORS

- See https://stackoverflow.com/questions/11933626/access-control-allow-origin-header-not-working-what-am-i-doing-wrong/11951532#11951532

### DIFF
--- a/src/supplemental/http/http_server.c
+++ b/src/supplemental/http/http_server.c
@@ -43,7 +43,9 @@ struct nng_http_handler {
 	bool            getbody;
 	void *          data;
 	nni_cb          dtor;
+	void *          res_mod_data;
 	void (*cb)(nni_aio *);
+	int (*res_mod_cb)(void *, nni_http_res *);
 };
 
 typedef struct http_sconn {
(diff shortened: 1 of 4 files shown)


robiwano:
This is a WIP, Access-Control-Allow-Origin should not be set per default. I have an idea for how to handle this properly.

src/supplemental/http/http_server.c
```diff
@@ -1496,6 +1496,8 @@ http_handle_dir(nni_aio *aio)
 
 	if (((rv = nni_http_res_alloc(&res)) != 0) ||
 	    ((rv = nni_http_res_set_status(res, NNG_HTTP_STATUS_OK)) != 0) ||
+	    ((rv = nni_http_res_set_header(
+	          res, "Access-Control-Allow-Origin", "*")) != 0) ||
```
gdamore: I think you probably already know this, but this is not a reasonable default.  I'm leaving this as a review, so that others can see why I'm not merging this this.

robiwano:
I opted for a more general solution, where it is possible to inject a callback precisely before http server response.

gdamore:
I will need to think about this before I approve it.  You should also update copyrights since at this point you're making some non-trivial changes to the code.

gdamore:
This will probably want some doc updates too.

robiwano:
Actually, I have an idea that makes this PR obsolete, and which is IMO the correct way to handle my "problem" (i.e. CORS). A new PR is coming up shortly.

PR #1158, opened 2020-01-20
Feature/uri length ordering

robiwano:
Adds possibility to have multiple handlers with same prefix, ordered in descending length. So you can serve a directory at "/" while having a GET handler serving "/rest-api". This makes sure that no CORS problems occur.

Also, this PR makes the nngcat bash scripts "executable" in git.

I concluded that no changes are needed to the nng_http_server_add_handler documentation, as what stands there is still applicable.

### DIFF
--- a/src/supplemental/http/http_server.c
+++ b/src/supplemental/http/http_server.c
@@ -3,6 +3,7 @@
 // Copyright 2018 Capitar IT Group BV <info@capitar.com>
 // Copyright 2018 QXSoftware <lh563566994@126.com>
 // Copyright 2019 Devolutions <info@devolutions.net>
+// Copyright 2020 Dirac Research <robert.bielik@dirac.com>
 //
 // This software is supplied under the terms of the MIT License, a
 // copy of which should be located in the distribution where this
@@ -37,6 +38,7 @@ struct nng_http_handler {
 	char *          method;
 	char *          host;
 	bool            tree;
+	bool            tree_exclusive;
 	nni_atomic_u64  ref;
 	nni_atomic_bool busy;
 	size_t          maxbody;
@@ -110,11 +112,12 @@ nni_http_handler_init(
 		return (NNG_ENOMEM);
 	}
 	NNI_LIST_NODE_INIT(&h->node);
-	h->cb      = cb;
-	h->data    = NULL;
-	h->dtor    = NULL;
-	h->host    = NULL;
-	h->tree    = false;
+	h->cb             = cb;
+	h->data           = NULL;
+	h->dtor           = NULL;
+	h->host           = NULL;
+	h->tree           = false;
+	h->tree_exclusive = false;
 	h->maxbody = 1024 * 1024; // By default we accept up to 1MB of body
 	h->getbody = true;
 	*hp        = h;
@@ -177,7 +180,19 @@ nni_http_handler_set_tree(nni_http_handler *h)
 	if (nni_atomic_get_bool(&h->busy) != 0) {
 		return (NNG_EBUSY);
 	}
-	h->tree = true;
+	h->tree           = true;
+	h->tree_exclusive = false;
+	return (0);
+}
+
+int
+nni_http_handler_set_tree_exclusive(nni_http_handler *h)
+{
+	if (nni_atomic_get_bool(&h->busy) != 0) {
+		return (NNG_EBUSY);
+	}
+	h->tree           = true;
+	h->tree_exclusive = true;
 	return (0);
 }
 
@@ -1115,8 +1130,8 @@ nni_http_server_add_handler(nni_http_server *s, nni_http_handler *h)
 	}
 
 	nni_mtx_lock(&s->mtx);
-	// General rule for finding a conflict is that if either string
-	// is a strict substring of the other, then we have a
+	// General rule for finding a conflict is that if either uri
+	// string is an exact duplicate of the other, then we have a
 	// collision.  (But only if the methods match, and the host
 	// matches.)  Note that a wild card host matches both.
 	NNI_LIST_FOREACH (&s->handlers, h2) {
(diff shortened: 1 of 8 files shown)


gdamore:
This sounds like a very good change.  I need to go over it carefully.  Thanks for the contribution.

gdamore:
FYI, I've been planning to rewrite the server and client test suites.  I do notice that you didn't update copyrights.  Was that intentional?  It seems like you're adding real value here, so you should get credit.

gdamore:
Having looked at this, there is a serious problem with the design.

Essentially, I think you've missed the point of the h->tree parameter, which is that you have the option of either registering for ownership of the entire tree, or for an exact URL.

This distinction is really important, because in some use cases (REST APIs) we actually want to put parameters on the URL line -- e.g. GET /object/<object-id>  -- we would have a handler for GET /object/ and then look at the full path information.

The idea of having a "fallback" handler, such that the tree logic is used, but we allow a more specific handler to be registered without considering it a conflict, is a good idea.

I think there is design work required.

src/supplemental/http/http_server.c
```diff
@@ -1146,26 +1146,31 @@ nni_http_server_add_handler(nni_http_server *s, nni_http_handler *h)
...
-		if (len2 > len) {
-			if ((h2->uri[len] == '/') && (h->tree)) {
-				nni_mtx_unlock(&s->mtx);
-				return (NNG_EADDRINUSE);
-			}
-		} else if (len > len2) {
-			if ((h->uri[len2] == '/') && (h2->tree)) {
-				nni_mtx_unlock(&s->mtx);
-				return (NNG_EADDRINUSE);
+		if (strncmp(h->uri, h2->uri, len) != 0) {
```
gdamore: Given the lengths are the same (above), you can just use strcmp(). here.

tests/httpserver.c
```diff
@@ -695,6 +695,46 @@ TestMain("HTTP Server", {
...
+
+		So(nng_http_handler_alloc_directory(&h, "/", workdir) == 0);
+		So(nng_http_server_add_handler(s, h) == NNG_EADDRINUSE);
+		nng_http_handler_free(h);
+
+		So(nng_http_handler_alloc_directory(&h, "/subdir", workdir) ==
+		    0);
+		So(nng_http_server_add_handler(s, h) == 0);
+
+		So(nng_http_handler_alloc_directory(&h, "/subdir", workdir) ==
```
gdamore: I'd really like to have a test to ensure that the handler for the more specific one is called when a more specific path is provided.

src/supplemental/http/http_server.c
```diff
@@ -1146,26 +1146,31 @@ nni_http_server_add_handler(nni_http_server *s, nni_http_handler *h)
 		while ((len2 > 0) && (h2->uri[len2 - 1] == '/')) {
 			len2--; // ignore trailing '/'
 		}
-		if (strncmp(h->uri, h2->uri, len > len2 ? len2 : len) != 0) {
-			continue; // prefixes don't match.
+
+		if (len != len2) {
+			continue; // length mismatch
```
gdamore: I am someone concerned by this change.  There is a semantic change on the URI handler, and while I think it is usually a good one, it won't necessarily always be so.

I think we might want a way to get the old matching behavior for h->tree handlers -- maybe another option.  Essentially I think it may be desirable for some use cases (think nanomsg websocket transports) where ownership of a subtree is "exclusive".

gdamore: Please think about whether you can achieve this.  Also the minor updates requested should be addressed -- and maybe copyright update?

robiwano: Yes, I'll look at this. Off the top of my head, shouldn't the longest uri match resolve this naturally? I.e. having a tree handler at "/tree" will handle "/tree/*", and one at "/" will resolve everything else. But since "/tree" handler lies before "/", the resolving is automagik. I'm not enough savvy about these technologies to see the problem this might incur.

tests/httpserver.c
robiwano: Yes, good point.

src/supplemental/http/http_server.c
robiwano: > Essentially, I think you've missed the point of the h->tree parameter, which is that you have the option of either registering for ownership of the entire tree, or for an exact URL.
> 
> This distinction is really important, because in some use cases (REST APIs) we actually want to put parameters on the URL line -- e.g. GET /object/ -- we would have a handler for GET /object/ and then look at the full path information.
> 

Ok, I see your point. And I now understand why I encountered the "address conflict" problem. In my C++ REST framework (https://github.com/robiwano/siesta) I want to be able to register REST handlers with regular expressions, which means I need to register a tree handler, and resolve the path wrt regexes in my handler callback. So basically, I registered multiple tree handlers which in the current nng implementation induces a conflict. I will address this problem in my framework instead so that if a full path without regexes are given, the handler is registered without the tree option. 

> The idea of having a "fallback" handler, such that the tree logic is used, but we allow a more specific handler to be registered without considering it a conflict, is a good idea.
> 
> I think there is design work required.

Yes, this would be something else.

src/supplemental/http/http_server.c
gdamore: My point is that someone registering at /tree might actually want to exclude a registration at /tree/subtree

This is really something that should be configurable.  I'm not disagreeing that your current implementation is not a bad default -- but it is a change, and we need a way to give the old behavior.  (In particular I think the ws:// API for websockets may kind of depend on it.)

robiwano: Something like:
`nng_http_handler_set_tree(nng_http_handler *h, int allow_sub_tree)` ?
or:
`nng_http_handler_set_tree_non_exclusive(nng_http_handler *h)` ?

gdamore: Well, set_tree() is an API that already exists so we don't want to change that.

But I could see having nng_http_handler_set_tree_exclusive() as a variant that gives the exclusive access version.

Probably this should be the default for the websocket transport for nanomsg, but not for other consumers of the HTTP framework.

robiwano: I'll give this a go.

robiwano:
The default behavior for a tree handler is non-exclusiveness, with the possibility to set it to exclusive with `nng_handler_set_tree_exclusive`. Thus I changed directory handler init to maintaining behavior. You mention websocket transport, however I couldn't find any users for `nng_handler_set_tree` in the codebase? I'm probably just missing it.

PR #1166, opened 2020-01-21
fixes #1162 req_test needs to have better retry tests

### LINKED ISSUE #1162: req_test needs to have better retry tests
The compat_reqrep test covers this, but we need the same coverage in the stock test.

In particular, closing a peer, re-connecting, should cause an immediate retry.   Likewise, if we wait for the timer to expire, we should be to see that we receive the the request twice on the REP side.

gdamore:
This actually uncovered a subtle problem with retry, where we could
wind up rescheduling the retry on the wrong pipe.  We are intended to
have tries be "sticky" (as long as the pipe that services the request
originally is still available.)

A subtle note though: when devices are involved, the request won't
necessarily be directed the same way, as the device doesn't have the
state to remember that.

contributing guidelines.

### DIFF
--- a/src/protocol/reqrep0/req_test.c
+++ b/src/protocol/reqrep0/req_test.c
@@ -184,6 +184,105 @@ test_req_rep_exchange(void)
 }
 
 void
+test_req_resend(void)
+{
+	nng_socket req;
+	nng_socket rep;
+
+	TEST_NNG_PASS(nng_req0_open(&req));
+	TEST_NNG_PASS(nng_rep0_open(&rep));
+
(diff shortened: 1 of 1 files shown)


gdamore:
I think I may want to back this change out -- the "sticky" routing for REQ is rather problematic...  its not going to be reliable in device networks, a stuck REP peer can wind up preventing forward progress, etc.

gdamore:
I rolled back the sticky part of this -- sticky req routing is a terrible idea.  I think some of the other implementations attempt to do that, but as I've thought about it, I think you don't ever really want that.

PR #1184, opened 2020-02-05
Fix possible use after free

ev-ermakov:


src/platform/posix/posix_resolv_gai.c
```diff
@@ -264,6 +264,7 @@ resolv_ip(const char *host, const char *serv, int passive, int family,
 	    sizeof(item->name_buf)) {
 		NNI_FREE_STRUCT(item);
 		nni_aio_finish_error(aio, NNG_EADDRINVAL);
+		return;
 	} else {
```
gdamore: Generally speaking I prefer to avoid the else if you're returning.  I'll follow up with that later.

src/core/message.c
```diff
@@ -490,6 +490,7 @@ nni_msg_alloc(nni_msg **mp, size_t sz)
 	if (rv != 0) {
 		nni_chunk_free(&m->m_header);
 		NNI_FREE_STRUCT(m);
+		return rv;
```
gdamore: Stylistically we always put return values in parens.

src/platform/posix/posix_resolv_gai.c
gdamore: Actually, maybe it would be best if you did that... as there is also another style nit.

gdamore: Thank you for your contribution.

There are two minor style nits that ideally I'd like you to resolve before I merge.  Best would be to resubmit (squashed and rebased) for a clean git history.

If you would rather not do that, then I can do it for you.  Please let me know within a day or so, because if I've not heard back then I'll probably just do it myself -- we definitely do want these changes.

src/core/message.c
```diff
@@ -490,6 +490,7 @@ nni_msg_alloc(nni_msg **mp, size_t sz)
 	if (rv != 0) {
 		nni_chunk_free(&m->m_header);
```
gdamore: Outside of the nits found, this looks good.

src/platform/posix/posix_resolv_gai.c
ev-ermakov: > Generally speaking I prefer to avoid the else if you're returning. I'll follow up with that later.

I hope I understand you correctly.

gdamore:
That one sanitizer build failure isn't your problem.  It's a bug  in the test code, and happens rarely.  (Race in the test code, not in the library.)

gdamore:
I'm rerunning that test to get the rest of the code covered.

gdamore:
Thanks for your contribution!

ev-ermakov:
Well, this looks like the tests have passed. Thanks :)

PR #1185, opened 2020-02-06
fixes #1180 use of inline strdup in a few places

### LINKED ISSUE #1180: use of inline strdup instances in a few places
We have some places where we do the logical equivalent of nni_strdup, but actually are doing alloc and strncpy or memcpy.

We should replace those with calls to nni_strdup() (and any free'ing with nni_strfree().)

This should help eliminate a few false positives from security scans.

gdamore:
While here we converted a couple of cases of local aio's from
heap to stack allocations, avoiding one of the error cases
associated with those.

contributing guidelines.


gdamore:
This analysis wrong.  I'm closing it for now.

PR #1193, opened 2020-02-08
Initial support for configuring TLS versions.

gdamore:
Docs forthcoming.  This should enable the use of older TLS v1.0
and 1.1, and in the future can enable v1.3.  Note that by default
we still force TLS v1.3.

A header definition for pluggable TLS engine is provided as well,
but this is a work in progress and not yet usable.


gdamore:
This was handled differently and the underlying capability is already integrated.

PR #1197, opened 2020-02-14
Fix error checks

tromgy:
Issue: starting the client(s) first, and the server afterwards blocks communication in these demo samples.

Most error codes in `nng_errno_enum` are positve values, so the existing error check would not catch something like `NNG_ECONNREFUSED`, the execution would proceed, and the communication would get blocked.

With this change, if the server is not able to receive, the client would exit with `fatal` instead of hanging.


gdamore:
Good catch!  Thanks!

PR #1199, opened 2020-02-15
Update comments in http.h

ev-ermakov:
fixes #

contributing guidelines.

### DIFF
--- a/include/nng/supplemental/http/http.h
+++ b/include/nng/supplemental/http/http.h
@@ -306,9 +306,9 @@ typedef struct nng_http_handler nng_http_handler;
 // first output using nng_aio_set_output.  If it does not do so, or supplies
 // NULL, then it must send a response itself.
 //
-// The callback should return 0 in most circumstances; if it returns anything
-// other than 0 then the connection is terminated (after possibly sending
-// a 500 error response to the client.)
+// The callback should complete with a result of 0 in most circumstances.
+// If it completes with an error, then the connection is terminated, after
+// possibly sending a 500 error response to the client.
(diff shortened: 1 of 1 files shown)


ev-ermakov:
```c
// The callback should return 0 in most circumstances; ...
```
Oh, maybe I misunderstood this comment. The callback returns void itself.

include/nng/supplemental/http/http.h
```diff
@@ -305,10 +305,6 @@ typedef struct nng_http_handler nng_http_handler;
 // Upon completion, the handler should store an nng_http_res * as the
 // first output using nng_aio_set_output.  If it does not do so, or supplies
 // NULL, then it must send a response itself.
-//
-// The callback should return 0 in most circumstances; if it returns anything
-// other than 0 then the connection is terminated (after possibly sending
-// a 500 error response to the client.)
```
gdamore: Why did you remove this comment.  I think it's accurate, no?

gdamore: I see the problem -- the issue is the word "return".
This should have have been reworded as:

"The callback should complete with a result of 0 most circumstances.  If it completes with an error,
then the connection is terminated, after possibly sending a 500 error response to the client"

gdamore: The critical information we need to convey is that a non-successful completion of the callback closes the connection (maybe returning a 500 error code as well.)

ev-ermakov: Agreed! I updated the pull request with the suggested changes.

PR #1207, opened 2020-02-24
fixes #1202 More than 120 threads was started by NNG

### LINKED ISSUE #1202: More than 120 threads was started by NNG
our contribution guidelines.

## NNG & Platform details.
1. OS: Ubuntu 18.04 LTS (x86-64)
2. IDE: CLion
3. NNG: 1.1.1
4. Compiler: GCC-5

When I break the app, I found more than 120 threads was started by NNG.

## Expected Behavior
Not so many threads.

## Actual Behavior
More than 120 threads was started by NNG.

## Steps to Reproduce
Just a simple NNG app.

gdamore:
This introduces a new CMake option, NNG_MAX_TASKQ_THREADS, with
a default value of 16.  The number of taskq workers will generally
be calculated as vcpu * 2.  This new value, if not zero, sets an
upper bound.  Note that the value should be at least two, in order
to ensure no deadlocks occur.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,6 +135,12 @@ if (NNG_NUM_TASKQ_THREADS)
 endif ()
 mark_as_advanced(NNG_NUM_TASKQ_THREADS)
 
+set(NNG_MAX_TASKQ_THREADS 16 CACHE STRING "Upper bound on taskq threads, 0 for no limit")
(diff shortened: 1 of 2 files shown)


gdamore:
Note that it will be ~impossible to cover this change in CI, because our CI environments only expose a single or maybe two cores.  I've verified this change on my system which has 12 cores and thus reduces the worker count to 16.  16 is an arbitrary number, but it's a starting point.

PR #1211, opened 2020-02-27
Inlineheader

gdamore:

contributing guidelines.


gdamore:
Initial performance comparisons are inconclusive (differences too small to draw any meaningful conclusions).

I think the biggest wins will be for systems where there are many threads, and the heap has more contention.  So this should help with scaling out.

PR #1233, opened 2020-04-13
fixes #1007 nngcat should accept data from stdin

### LINKED ISSUE #1007: nngcat should accept --data or --file over stdin
our contribution guidelines.

## NNG & Platform details.

nng 1.1.1 (via FreeBSD ports)

## Expected Behavior

```
ping localhost | nngcat --pub --listen=tcp://127.0.0.1:4567 -i 1 -v --file -
```

## Actual Behavior

```
Protocol requires either --file or --data.
```

it seems nngcat doesn't accept piped stdin, nor via `--file -` or `--data -` or the same without `-`.

## Steps to Reproduce

See above, run `nngcat --sub --dial=tcp://127.0.0.1:4567 --raw` in another term.

ehaas:
Allows passing a filename of `-` to the `--file` option in order to read from
stdin. Also removes the requirement for the file to be seekable - this allows
nngcat to be used as part of a shell pipeline, or with other non-seekable files


tools/nngcat/nngcat.c
```diff
@@ -309,28 +310,49 @@ static void
...
+		}
+		if (feof(f)) {
+			break;
+		}
+		if (total_read == allocation_size) {
+			if (allocation_size > SIZE_MAX / 2) {
+				fatal("Out of memory.");
+			}
+			allocation_size *= 2;
+			if ((fdata = realloc(fdata, allocation_size + 1)) == NULL) {
```
ehaas, line 345: Codacy complains that this is a memory leak, but since we just `fatal` on realloc failure anyway I don't think it matters too much. Happy to fix it if you like though.

tools/nngcat/nngcat_stdin_pipe_test.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
dch, line 1: can we use `#!/usr/bin/env bash` here to facilitate testing on a wider group of unix variants?

ehaas: I was just following the example in the other nngcat test files (e.g. `nngcat_pubsub_test.h`) - probable worth running it by @gdamore

tools/nngcat/nngcat.c
gdamore: I'd like to fix it please.  I agree it's probably benign, but some users get unhappy when automated code analysis tools trigger complaints.

tools/nngcat/nngcat_stdin_pipe_test.sh
gdamore: Yes, I think its probably a good idea to do this.  We should probably fix the others too.

Really, I'd like to move away from using a shell script to drive these tests, and I have a plan to do that at some point.

tools/nngcat/nngcat.c
ehaas: Should be fixed now

gdamore:
Cool.  Can you squash these for me?
I'm going to tag 1.3.1 for release today.  Probably will *not* include this, since it represents new functionality, but instead will make this one of the first commits *after* I tag 1.3.1.  (One of the first for 1.4.)

ehaas:
Rebased against master, squashed, and added a usage note.

ehaas:
It seems like some unrelated tests are failing. In a previous build I also saw an error from `tests/compat_ws.c` (unfortunately I don't have any other info, except that it failed on the Linux build).

gdamore:
Those errors are unrelated.

Furthermore, I've decided that your changes fall below the threshold of feature, and represent really bug fixes.  So I'm going to integrate in 1.3.1.  Thanks for your contribution.

PR #1238, opened 2020-05-07
fixes #1236 Deadlock triggered on nng_close

### LINKED ISSUE #1236: Deadlock triggered on nng_close
Creator of [pynng](https://github.com/nanomsg/pynng) here.  Under certain circumstances our unit tests were triggering a deadlock when closing sockets, and I was able to eventually reproduce it in plain C.  I think the key point here is that 
```c
#include <assert.h>
#include <stdio.h>
#include <stdlib.h>

#include <nng/nng.h>
#include <nng/protocol/pair0/pair.h>

#define CHECK(x) do { \
    int ret = (x); \
    if (ret != 0) { \
        printf("error %d: %s\n", ret, nng_strerror(ret));\
        abort(); \
    } \
} while (0)

char addr[] = "tcp://localhost:9899";
int main() {
    char buf[50];
    for (int i = 0; ; i++) {
        size_t got = sizeof buf;
        nng_socket s0, s1;
        CHECK(nng_pair_open(&s0));
        CHECK(nng_listen(s0, addr, NULL, 0));
        CHECK(nng_pair_open(&s1));
        CHECK(nng_dial(s1, addr, NULL, 0));

        CHECK(nng_send(s0, "hello", 5, 0));
        CHECK(nng_recv(s1, buf, &got, 0));
        printf("%d\n", i);
        nng_close(s1);
        nng_close(s0);
    }

    return 0;
}

```

This is likely the same issue as #1219.

----------------

It seems relevant that the deadlock **does not happen** if the `nng_send` and `nng_recv` calls are removed.

gdamore:
I'm not sure if this actually closes issue or not.  It may also have other problems.  I'd like folks to help me test it out, ideally.

### DIFF
--- a/src/core/aio.c
+++ b/src/core/aio.c
@@ -328,6 +328,7 @@ nni_aio_schedule(nni_aio *aio, nni_aio_cancelfn cancelfn, void *data)
 		// Convert the relative timeout to an absolute timeout.
 		switch (aio->a_timeout) {
 		case NNG_DURATION_ZERO:
+			nni_task_abort(&aio->a_task);
 			return (NNG_ETIMEDOUT);
 		case NNG_DURATION_INFINITE:
 		case NNG_DURATION_DEFAULT:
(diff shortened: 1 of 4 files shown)


gdamore:
I've stopped using circleci.  Should be better now.

gdamore:
Another change -- I think also related to 1236 - added.  Please retest.

PR #1248, opened 2020-05-26
fix typo in nng_cv_wait example

dehorsley:
`cv_wake` should be `nng_cv_wake`


gdamore:
Typo is not quite the right words here.  Oversight is more accurate.  Thanks for your contribution.

PR #1257, opened 2020-06-16
fix #1252

### LINKED ISSUE #1252: nng_msg_clear also clears the header
[`nng_msg_clear` docs](https://nng.nanomsg.org/man/v1.3.0/nng_msg_clear.3.html) state that it only clears the message body, but in 1.3.0 it clears both the body and header.  This was not the case in earlier releases.

## NNG & Platform details.

1.3.0

## Expected Behavior

`nng_msg_clear` only clears the message body and leaves the header untouched.

## Actual Behavior

`nni_msg_clear` now also clears the now inlined header from this CL (https://github.com/nanomsg/nng/commit/abab1392c11395d387e3072c4f5788d284846093#diff-f719e7b45a41a7b3f01eea3660f52434).

jeikabu:
- nng_msg_clear should only clear the message body and not the head.  Add test to verify

fixes #1252

### DIFF
--- a/src/core/message.c
+++ b/src/core/message.c
@@ -578,7 +578,6 @@ nni_msg_header_append_u32(nni_msg *m, uint32_t val)
 void
 nni_msg_clear(nni_msg *m)
 {
-	m->m_header_len = 0;
 	nni_chunk_clear(&m->m_body);
 }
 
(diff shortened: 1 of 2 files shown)


gdamore:
Thanks for the contribution!

PR #1259, opened 2020-07-06
Print NNG configuring message to STATUS mode.

ScatteredRay:
fixes printing to stderr causing other build tools to detect a build error.

The default mode for "message()" under default configuration prints to stderr, some build tools take this to mean an error has occurred incorrectly.


gdamore:
Thanks for your contribution.

PR #1261, opened 2020-07-09
Add a nng_stream example paired with a socket server on win/linux

hugolm84:
This PR adds a nng_stream example that connects to an arbitrary socket. Either start a new server with `./stream -s <portnumber>` or connect to an existing one using `./stream -c <url>`

### DIFF
--- a/demo/stream/platform/posix/server.c
+++ b/demo/stream/platform/posix/server.c
@@ -1,0 +1,53 @@
+// Copyright 2020 Hugo Lindström <hugolm84@gmail.com>
+
+// This software is supplied under the terms of the MIT License, a
+// copy of which should be located in the distribution where this
+// file was obtained (LICENSE.txt).  A copy of the license may also be
+// found online at https://opensource.org/licenses/MIT.
+
+#include <netinet/in.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <strings.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
+void
+error(const char *msg)
+{
+	perror(msg);
+	exit(1);
+}
+
+int
+server(int portno)
+{
+	int                sockfd, newsockfd;
+	socklen_t          clilen;
+	struct sockaddr_in serv_addr, cli_addr;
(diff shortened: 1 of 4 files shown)


gdamore:
Thanks for your contribution.  I'm going to review this closely before I integrate.

hugolm84:
> Thanks for your contribution. I'm going to review this closely before I integrate.

Hi, I understand that this simple demo might not be compelling enough, let me know if it needs to be expanded.

gdamore:
I just hadn't got a chance to look at it yet.

demo/stream/platform/posix/server.c
```diff
@@ -0,0 +1,53 @@
+// Copyright 2020 Hugo Lindström <hugolm84@gmail.com>
+
+// This software is supplied under the terms of the MIT License, a
+// copy of which should be located in the distribution where this
+// file was obtained (LICENSE.txt).  A copy of the license may also be
+// found online at https://opensource.org/licenses/MIT.
```
gdamore, line 6: I think you should make the server once, and use the STREAM APIs.  I realize part of what you might want to show is that these work with ordinary TCP/IP applications, but we want to show that the STREAMs API can be used on both the server and the client side.

gdamore: If you don't have time for that, let me know, and I'll work on it.

hugolm84: Yes, my point was to show that nng can be used even with legacy applications that uses TCP/IP over a standard OS socket. I did not think of showcasing that nng_stream can be used as a server as well. Perhaps showcasing both have its own value? In any case, I can make some time to implement the server with STREAM.

gdamore: I think showcasing both has value.  Plus that way its just one program not two for the server.  Showing the portability benefit is good. :-)

gdamore:
I'm just going to merge this for now.  It's a demo.  It doesn't have to be perfect.  Once it's integrated we can further improve it.

PR #1266, opened 2020-07-26
fixes #844 WebSocket wildcard host errors

### LINKED ISSUE #844: WebSocket WebSocket wildcard host errors
our contribution guidelines.

## NNG & Platform details.

nng-1.1.1, Debian GNU/Linux 9.5 (Stretch)

## Expected Behavior

When listening to `ws://*:45454/foobar` I would expect that : 

1. `nng_dial` to `ws://localhost:45454/foobar` works
2. that this listen does on both IPv4 and IPv6 on localhost, and the same for external/LAN interface

The documentation, indeed seems to explain it's dependent of the underlying system but as I think it's a common setup for many OS, and that it works out-of-the-box on macOS, i'm wondering if I'm missing something.

>  To listen to all ports on the system, the host name may be elided from the URL on the listener. This will wind up listening to all interfaces on the system, with possible caveats for IPv4 and IPv6 depending on what the underlying system supports. (On most modern systems it will map to the special IPv6 address ::, and both IPv4 and IPv6 connections will be permitted, with IPv4 addresses mapped to IPv6 addresses.)

## Actual Behavior

What happens is the `listen` is done only on IPv4 and the `dial` is using IPv6.

1. Is there a way to specify that you only want IPv4 or IPv6 ?
2. As IPv6 is now the preferred protocol and I would prefer a more transparent approach, is there something that can be done to avoid specifying/limit the IP protocol version ?
3. It's an almost stock installation of Debian 9 but maybe the setup is incorrect and some configuration should be done ?

### LINKED ISSUE #1224: wss fails on IPV6 address
wss fails to work on IPV6 addresses. No problem is seen with IPV4 addresses and with wss://localhost:port.

The attached file demonstrates this issue on Linux.

tls or tls+tcp just works in the attached file.

Please advise if wss is expected to work with IPV6 addresses.

Regards.

[nng_wss_ipv6_c.cpp.txt](https://github.com/nanomsg/nng/files/4355165/nng_wss_ipv6_c.cpp.txt)

gdamore:
fixes #1224 wss fails on IPV6 address

This fixes bugs and inconsistencies in the way addresses are
handled for HTTP (and consequently websocket).  The Host:
address line needs to look at numeric IPs and treat wildcards
as if they are not specified, and needs to understand the IPv6
address format using brackets (e.g. [::1]:80).

### DIFF
--- a/src/core/platform.h
+++ b/src/core/platform.h
@@ -339,6 +339,12 @@ extern void nni_tcp_resolv(const char *, const char *, int, int, nni_aio *);
 // service names using UDP.
 extern void nni_udp_resolv(const char *, const char *, int, int, nni_aio *);
 
+// nni_parse_ip parses an IP address, without a port.
+extern int nni_parse_ip(const char *, nng_sockaddr *);
+
+// nni_parse_ip_port parses an IP address with an optional port appended.
+extern int nni_parse_ip_port(const char *, nng_sockaddr *);
+
 //
 // IPC (UNIX Domain Sockets & Named Pipes) Support.
 //
--- a/src/platform/posix/posix_resolv_gai.c
+++ b/src/platform/posix/posix_resolv_gai.c
@@ -1,5 +1,5 @@
 //
-// Copyright 2019 Staysail Systems, Inc. <info@staysail.tech>
+// Copyright 2020 Staysail Systems, Inc. <info@staysail.tech>
 // Copyright 2018 Capitar IT Group BV <info@capitar.com>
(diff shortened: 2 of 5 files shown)


gdamore:
Note also this uses IP address parsing to ensure that we cope with the various different yet equivalent ways of handling IP addresses.  However v4mapped addresses might not behave properly.

gdamore:
The code complexity stuff is annoying.  There really isn't a great way to make that better.

PR #1286, opened 2020-08-13
Fixes for MSVC1800

hugolm84:
Currently, `nng` does not compile with Visual Studio 12.0 2013 (MSVC 1800). This PR attempts to provide a fix for that.

Added optional `nng_snprintf` function, could later be the ONE function used through out, replacing snprintf, _snprintf
Renamed the NNI_PUT macro variable from `u` to `n`, as it clashes with the number suffix.

Discussion: Not sure `nng.h` is the best header to include `nng_snprintf.h` from. Happy to move it elsewhere should you have a better suggestion.

### DIFF
--- a/include/nng/supplemental/util/nng_snprintf.h
+++ b/include/nng/supplemental/util/nng_snprintf.h
@@ -1,0 +1,42 @@
+#ifndef NNG_SNPRINTF_H
+#define NNG_SNPRINTF_H
+
+#include <stdarg.h>  // NOLINT
+#include <stdio.h>   // NOLINT
+
+#if !defined(__cplusplus) && defined(_MSC_VER)
+#define inline __inline
+#endif  //__cplusplus && _MSC_VER
+
+#ifdef __cplusplus
+extern "C" {
+#endif  //__cplusplus
+
+#if !defined(_TRUNCATE)
+#define _TRUNCATE ((size_t)-1)
+#endif  //_TRUNCATE
+
+inline int nng_snprintf(char* const _Buffer, int const _BufferCount, char const* const _Format, ...) {
+  int ret = 0;
+  va_list args;
+  va_start(args, _Format);
+#if _MSC_VER
+  ret = vsnprintf_s(_Buffer, _BufferCount, _TRUNCATE, _Format, args);
+#else
+  ret = vsnprintf(_Buffer, _BufferCount, _Format, args);
+#endif
+  va_end(args);
+  return ret;
+}
+
+#ifdef USE_NNG_SNPRINTF
+#if USE_NNG_SNPRINTF
+#define snprintf nng_snprintf
+#endif
+#endif
+
+#ifdef __cplusplus
+}
+#endif  //__cplusplus
+
+#endif  // NNG_SNPRINTF_H
--- a/src/core/defs.h
+++ b/src/core/defs.h
@@ -85,30 +85,30 @@ typedef struct {
 #define NNI_ALLOC_STRUCTS(s, n) nni_zalloc(sizeof(*s) * n)
 #define NNI_FREE_STRUCTS(s, n) nni_free(s, sizeof(*s) * n)
 
-#define NNI_PUT16(ptr, u)                                    \
+#define NNI_PUT16(ptr, n)                                    \
 	do {                                                 \
-		(ptr)[0] = (uint8_t)(((uint16_t)(u)) >> 8u); \
-		(ptr)[1] = (uint8_t)((uint16_t)(u));         \
+		(ptr)[0] = (uint8_t)(((uint16_t)(n)) >> 8u); \
+		(ptr)[1] = (uint8_t)((uint16_t)(n));         \
 	} while (0)
 
-#define NNI_PUT32(ptr, u)                                     \
+#define NNI_PUT32(ptr, n)                                     \
 	do {                                                  \
-		(ptr)[0] = (uint8_t)(((uint32_t)(u)) >> 24u); \
-		(ptr)[1] = (uint8_t)(((uint32_t)(u)) >> 16u); \
-		(ptr)[2] = (uint8_t)(((uint32_t)(u)) >> 8u);  \
-		(ptr)[3] = (uint8_t)((uint32_t)(u));          \
+		(ptr)[0] = (uint8_t)(((uint32_t)(n)) >> 24u); \
+		(ptr)[1] = (uint8_t)(((uint32_t)(n)) >> 16u); \
+		(ptr)[2] = (uint8_t)(((uint32_t)(n)) >> 8u);  \
+		(ptr)[3] = (uint8_t)((uint32_t)(n));          \
 	} while (0)
 
-#define NNI_PUT64(ptr, u)                                     \
+#define NNI_PUT64(ptr, n)                                     \
 	do {                                                  \
-		(ptr)[0] = (uint8_t)(((uint64_t)(u)) >> 56u); \
-		(ptr)[1] = (uint8_t)(((uint64_t)(u)) >> 48u); \
-		(ptr)[2] = (uint8_t)(((uint64_t)(u)) >> 40u); \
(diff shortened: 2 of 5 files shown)


include/nng/supplemental/util/nng_snprintf.h
```diff
@@ -0,0 +1,42 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif  //__cplusplus
+
+#if !defined(_TRUNCATE)
+#define _TRUNCATE ((size_t)-1)
+#endif  //_TRUNCATE
+
+inline int nng_snprintf(char* const _Buffer, int const _BufferCount, char const* const _Format, ...) {
```
gdamore, line 19: This by the way doesn't conform to our other coding style (see clang-format).  Having said that, we also don't make use if inline anywhere else in nng today.  I've been cautious about doing thus far -- no where this function is used is sufficiently performance sensitive to justify the effort the stack frame any way.

src/core/defs.h
```diff
@@ -85,30 +85,30 @@ typedef struct {
 #define NNI_ALLOC_STRUCTS(s, n) nni_zalloc(sizeof(*s) * n)
 #define NNI_FREE_STRUCTS(s, n) nni_free(s, sizeof(*s) * n)
 
-#define NNI_PUT16(ptr, u)                                    \
```
gdamore, line 88: I don't really understand the motivation for this change.  The use of "u" as a suffix is unambiguous with this (it always is prefixed by numeric digits!) -- if this was causing some compiler grief I'd like to see compiler output, because it strongly indicates a compiler defect (and a serious one at that.)

include/nng/supplemental/util/nng_snprintf.h
gdamore: The fact that you have to redefine it as __inline for MSVC is a telling statement btw.  I don't consider inline portable.

gdamore:
So I'm of a mixed mind about this PR in general.  It seems like snprintf() exists pretty much everywhere, but it breaks on some older version of MSVC.  Do we need to support that older version of Visual Studio, and if so why?  Really, folks should be building with VC 2015 or better which has a real snprintf().

If we're going to really fix this to add support for older older versions of MSVC, then we should actually provide code that actually checks the buffer and null terminates properly -- that's a security consideration.

As it is, I don't think just wrapping _snprintf() is a suitable approach.

include/nng/supplemental/util/nng_snprintf.h
```diff
@@ -0,0 +1,42 @@
...
+  int ret = 0;
+  va_list args;
+  va_start(args, _Format);
+#if _MSC_VER
+  ret = vsnprintf_s(_Buffer, _BufferCount, _TRUNCATE, _Format, args);
```
gdamore, line 24: I see -- this does the right thing to terminate safely.

I still am not thrilled with doing this work to support older MSVC -- there are three newer major releases of MSVC that would work without needing (2015, 2017, and 2019).

Exposing this functionality here in this way isn't right either.  If we want to make it a public function (and I'm not sure we do -- I don't intend NNG to become the universal platform portability library), we should stick a prototype in supplemental/util/platform.h and put MSVC implementation specific details in the win32 port layer.  Ideally it would look at the MSC version and only provide a builtin wrapper around _vsnprintf_s if the version was too old.

gdamore:
One other thought:  snprintf() is required per C99.  NNG requires C99 as a minimum requirement.  We make a special exception for Windows, because Windows doesn't offer a standard compliant C environment, Microsoft having decided long ago that it only wants to support C++ -- although even Microsoft has added some C99 stuff (like snprintf() back in MSVC 2015).

Hence the only change I'd really be willing to accept at this point would be a fix for older MSVC, that was isolated in the porting layer.  That change would provide an implementation of snprintf, not an implementation of some other function, which would only be exposed internally in the library.  And to be honest, I'm not even terribly happy about even that -- developers really should just use modern toolchains.  (The C99 specification is 21 years old, for cryin' out loud!)

src/core/defs.h
hugolm84: Hi, yes, this is not seen in any compiler newer than MSVC 1800.

```
cmake -G "Visual Studio 12 2013" -A Win32 ..
```
Error (one example):
```
  tls.c
C:\github\nng\src\transport\tls\tls.c(584): error C2059: syntax error : 'bad suffix on number' [C:\github\nng\build\src
\nng.vcxproj]
C:\github\nng\src\transport\tls\tls.c(584): error C2146: syntax error : missing ')' before identifier 'p' [C:\github\nn
g\build\src\nng.vcxproj]
C:\github\nng\src\transport\tls\tls.c(584): error C2059: syntax error : ')' [C:\github\nng\build\src\nng.vcxproj]
C:\github\nng\src\transport\tls\tls.c(585): warning C4293: '>>' : shift count negative or too big, undefined behavior [
C:\github\nng\build\src\nng.vcxproj]
```

Fixed by not using `u` as macro parameter.

include/nng/supplemental/util/nng_snprintf.h
hugolm84: If you like, I can happily add a source file instead of inline.

hugolm84:
> So I'm of a mixed mind about this PR in general. It seems like snprintf() exists pretty much everywhere, but it breaks on some older version of MSVC. Do we need to support that older version of Visual Studio, and if so why? Really, folks should be building with VC 2015 or better which has a real snprintf().
> 
> If we're going to really fix this to add support for older older versions of MSVC, then we should actually provide code that actually checks the buffer and null terminates properly -- that's a security consideration.
> 
> As it is, I don't think just wrapping _snprintf() is a suitable approach.

In light of newer comments, is this comment still valid?

include/nng/supplemental/util/nng_snprintf.h
hugolm84: I understand that due to resource constraints and over all sanity, projects have a need to limit it's support for older platforms. However, some projects that depend on `nng` have limitations on what compiler they can use. For one example, WinCE uses a hybrid compiler between VS 2012 and VS 2013. That compiler version will never be updated, so thats what we have to work with. Supporting VS13 is a step towards supporting WinCE.

In an ideal world, everyone would be using the latest and greatest, however, sometimes that's not possible. :/

Going forward, I like they suggestion of moving this in to the platform layer instead.

hugolm84:
> One other thought: snprintf() is required per C99. NNG requires C99 as a minimum requirement. We make a special exception for Windows, because Windows doesn't offer a standard compliant C environment, Microsoft having decided long ago that it only wants to support C++ -- although even Microsoft has added some C99 stuff (like snprintf() back in MSVC 2015).
> 
> Hence the only change I'd really be willing to accept at this point would be a fix for older MSVC, that was isolated in the porting layer. That change would provide an implementation of snprintf, not an implementation of some other function, which would only be exposed internally in the library. And to be honest, I'm not even terribly happy about even that -- developers really should just use modern toolchains. (The C99 specification is 21 years old, for cryin' out loud!)

Yes, adding `snprintf` et al for MSVC <= 1800 in the platform layer is something I could do instead. I understand its not an ideal situation but it would be neat if we could support MSVC. This is (seemingly) the only change (currently) needed to support older versions.

src/core/defs.h
gdamore: Wow.  That's shocking.  Completely and totally broken behavior.  I guess it also means you can't have a variable named "u".

Please submit this as a separate PR.

gdamore:
I'd say at this point, submit the fix for "u" being magical in a separate PR.  Then make the snprintf wrapper in the Win32 port layer, but conditionalize it on MSVC < 1900.

src/core/defs.h
hugolm84: Looking in to it a bit more, suffixing with upper case `U` works together with a lowercase macro variable named lower case `u`.

As `Letters in the integer literals are case-insensitive`, I'll make a separate PR with `u`'s and `U`'s.

hugolm84:
Closing in favor of separate PR's

PR #1291, opened 2020-08-18
MSVC1800 does not like mixing macro variable names and suffixes

hugolm84:
MSVC1800 does not like mixing lowercase macro variable named `u` together with the integral suffix lower case `u`. 
To fix this, lets use lower case for macro var and uppercase suffix instead.

Error: C:\github\nng\src\transport\tls\tls.c(584): error C2059: syntax error : 'bad suffix on number' [C:\github\nng\build\src\nng.vcxproj]

### DIFF
--- a/src/core/defs.h
+++ b/src/core/defs.h
@@ -87,27 +87,27 @@ typedef struct {
 
 #define NNI_PUT16(ptr, u)                                    \
 	do {                                                 \
-		(ptr)[0] = (uint8_t)(((uint16_t)(u)) >> 8u); \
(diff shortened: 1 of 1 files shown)


gdamore:
I think it would have been better to use "n" rather than "u" here.  I'm not sure I'll remember to capitalize one way or the other in the future, so your change to use "n" for this is a bit ... safer.

I still can't believe that MSVC botched this as badly as they did.

hugolm84:
This will not be accepted and implementation needs to be discussed elsewhere.

PR #1292, opened 2020-08-18
Add snprintf wrapper if MSC_VER is less than 1900

hugolm84:
`snprintf` does not exist in MSVC 1800 and below. This PR adds a wrapper around `vsnprintf_s` instead.

Unfortunately, this change also makes the `snprintf` function an internal implementation only available at linking, generating a compile warning:

`warning C4013: 'snprintf' undefined; assuming extern returning int`

To solve this, the files

```
src/supplemental/base64/base64_test.c
tests/convey.h
tests/testutil.h
tools/nngcat/nngcat.c
```

also needs to declare `snprintf` as extern, leading to code duplication throughout the test targets. 

One obvious way forward, to me, seems to be to add the declaration in

`nng/supplemental/util/platform.h`
 
instead. This would of course pollute the public header and may not be accepted.

### DIFF
--- a/src/platform/windows/win_snprintf.c
+++ b/src/platform/windows/win_snprintf.c
@@ -1,0 +1,34 @@
+#ifndef NNG_WIN_SNPRINTF_H
+#define NNG_WIN_SNPRINTF_H
+
+#if !NNG_HAVE_SNPRINTF
+#if defined(_MSC_VER) && (_MSC_VER < 1900)
+
+#include <stdarg.h> // NOLINT
+#include <stdio.h>  // NOLINT
+
+#if !defined(_TRUNCATE)
+#define _TRUNCATE ((size_t) -1)
+#endif //_TRUNCATE
+
+// Each of these functions takes a pointer to an argument list, then formats
+// and writes up to count characters of the given data to the memory pointed to
+// by buffer and appends a terminating null. If count is _TRUNCATE, then these
+// functions write as much of the string as will fit in buffer while leaving
+// room for a terminating null. If the entire string (with terminating null)
+// fits in buffer, then these functions return the number of characters written
+// (not including the terminating null); otherwise, these functions return -1
+// to indicate that truncation occurred.
+int
+snprintf(char *buffer, size_t count, const char *format, ...)
+{
+	va_list args;
+	va_start(args, format);
+	int ret = vsnprintf_s(buffer, count, _TRUNCATE, format, args);
+	va_end(args);
+	return ret;
+}
+
+#endif // _MSC_VER < 1900
+#endif // !NNG_HAVE_SNPRINTF
+#endif NNG_WIN_SNPRINTF_H
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -290,7 +290,10 @@ if (WIN32)
     list(APPEND NNG_LIBS ws2_32 mswsock advapi32)
     nng_check_sym(InitializeConditionVariable windows.h NNG_HAVE_CONDVAR)
     nng_check_sym(snprintf stdio.h NNG_HAVE_SNPRINTF)
-    if (NOT NNG_HAVE_CONDVAR OR NOT NNG_HAVE_SNPRINTF)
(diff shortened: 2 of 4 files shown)


gdamore:
Admittedly I still hate that we're fixing this.  I feel like we're adding technical debt to solve a problem that isn't really ours at all, and is simply resolved by using more modern tools.  Remind me again why people can't update their toolchains to a product that has been available for over 5 years?)

(For the record, when people asked about ports to XP, and even Vista, I've told them to go pound sand -- unless someone is paying for it, I can't afford effort to support every ancient system.  Especially when I can't even install them myself anymore.  I'm not sure if MSVC 2013 is still available even.)

Instead of trying to sort out the symbol problems (and I expect there will be a mess of them with 3rd party programs that may have their own snprintf implementations), I think a better approach would be to have windows have an "nni_win_snprintf()" in the port code, then supply a macro for snprintf in nng_impl.h.  The linker will see only nni_win_snprintf symbol names, and third party programs won't see our macro when it is tucked away there.

src/platform/windows/win_snprintf.c
```diff
@@ -0,0 +1,34 @@
+#ifndef NNG_WIN_SNPRINTF_H
+#define NNG_WIN_SNPRINTF_H
+
+#if !NNG_HAVE_SNPRINTF
+#if defined(_MSC_VER) && (_MSC_VER < 1900)
+
+#include <stdarg.h> // NOLINT
+#include <stdio.h>  // NOLINT
+
+#if !defined(_TRUNCATE)
```
gdamore, line 10: This is truly disgusting.  If the tool chain *also* doesn't support _TRUNCATE we must be *really* old?

gdamore: Also, is the value of that symbol documented this way?

gdamore: Oh, I see, this is probably MinGW-32... No thanks.  Explicitly we don't support mingw-32, and never will.  Mingw-64 is fine.  (mingw-32 lacks some APIs we depend on, and it would be non-trivial to fix those gaps without leaving unacceptable race conditions in the code.)

hugolm84:
> Admittedly I still hate that we're fixing this. I feel like we're adding technical debt to solve a problem that isn't really ours at all, and is simply resolved by using more modern tools. Remind me again why people can't update their toolchains to a product that has been available for over 5 years?)
> 
> (For the record, when people asked about ports to XP, and even Vista, I've told them to go pound sand -- unless someone is paying for it, I can't afford effort to support every ancient system. Especially when I can't even install them myself anymore. I'm not sure if MSVC 2013 is still available even.)
> 
> Instead of trying to sort out the symbol problems (and I expect there will be a mess of them with 3rd party programs that may have their own snprintf implementations), I think a better approach would be to have windows have an "nni_win_snprintf()" in the port code, then supply a macro for snprintf in nng_impl.h. The linker will see only nni_win_snprintf symbol names, and third party programs won't see our macro when it is tucked away there.

Hi, I was under the impression from previous discussion that a `nni_win_snprintf` method was not acceptable and that a drop-in replacement of the missing `snprintf` was preferred. If `nni_win_snprintf` is the way forward, I will change it. 

You can download and install Visual Studio´s down to 2010 from here https://visualstudio.microsoft.com/vs/older-downloads/

I understand that supporting old platforms can be and often is very taxing. I am happy to provide my efforts to make `nng` support VS13 and hope that those efforts can be merged upstream.

There can be many reasons as to why developers are stuck with a certain toolchain, for me, it is because my embedded devices still have hardware that runs applications on WinCE. The closest we can come in toolchain similarities with WinCE is VS2013, which makes development a lot easier: Following certain rules usually means that if tests runs green on the host, it usually goes green on the target.

src/platform/windows/win_snprintf.c
hugolm84: I will remove this.

hugolm84:
This will not be accepted and implementation needs to be discussed elsewhere.

PR #1310, opened 2020-10-30
fixes #1022 Support tcp6 link local addresses

### LINKED ISSUE #1022: Support tcp6 link local addresses
In order to properly use link local adresses the device needs to be specified. I propose the syntax detailed in this example:
`tcp6://[fe80::2%lan]:80 `

For example Linux ping supports this syntax:
`ping fe80::2%lan`

I did not find any issue discussing this or any way to use link local addresses with nng. If there are please tell me how to do it.

gdamore:
This is a test draft -- not really tested yet, since I don't have much in the way of systems to test with.  I'm just passing through the values provided by getaddrinfo().

Testing help is requested.

### DIFF
--- a/docs/man/nng_sockaddr_in6.5.adoc
+++ b/docs/man/nng_sockaddr_in6.5.adoc
@@ -27,6 +27,7 @@ typedef struct {
     uint16_t sa_family;
     uint16_t sa_port;
     uint8_t  sa_addr[16];
+    uint32_t sa_scope;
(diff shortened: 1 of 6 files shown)


gdamore:
FYI, I plan to merge this in -- even without actual formal testing, since testing this would be platform specific and I don't have the specifics on any one particular platform -- and also we have no support for IPv6 in any of our CI pipelines (pretty sure).

PR #1348, opened 2020-11-16
fixes #1071 tran_chkopt can be cleaned up

### LINKED ISSUE #1071: tran_chkopt can be cleaned up

gdamore:
This is a sweeping cleanup of the transport logic around options,
and also harmonizes the names used when setting or getting options.
Additionally, legacy methods are now moved into a separate file and
can be elided via CMake or a preprocessor define.

Fundamentally, the ability to set to transport options via the socket
is deprecated; there are numerous problems with this and my earlier
approaches to deal with this have been somewhat misguided.  Further
these approaches will not work with future protocol work that is
planned (were some options need to be negotiated with peers at the
time of connection establishment.)

Documentation has been updated to reflect this.  The test suites still
make rather broad use of the older APIs, and will be converted later.

contributing guidelines.


gdamore:
One note: This PR is a big step towards stabilizing the transport API.  There are developers that would like to be able to extend NNG with new transports, and having a clean API for that work is important -- the checkopt hack was contraindicated in this regard.

PR #1349, opened 2020-11-16
fixes #1347 windows IPC accept cancellation not wired up

### LINKED ISSUE #1347: windows IPC accept cancellation not wired up
Looking closely at the Windows IPC code, it appears that we never properly handled cancellation of an accept.  It also looks like we never wired nni_aio_schedule().

gdamore:

contributing guidelines.


gdamore:
This seems to be failing in CI/CD.

I think its possibly different timing causing a race against the IPC socket name combined with a naive test suite that reuses the same name.  I've also noticed that there have been other timeouts lately in GitHub particularly with Windows and Darwin targets.  (Linux targets seem fine.)

PR #1375, opened 2020-12-19
Fix compilation error

unspecd:
Fix possible compilation error.


gdamore:
Well, that's a nice catch -- but I'm astonished you ran into this -- because right now the only platforms NNG supports are Windows and those that declare POSIX. 

What platform did you hit this on, or was it found by inspection?

unspecd:
Just experimenting. :)
I am trying to add support for FreeRTOS.

PR #1383, opened 2020-12-23
call getrandom like the syscall it is, avoiding linking issues on som…

marijnfs:
getrandom was called like a function, causing linking issues on my system. This change at least avoids issues on my OpenSuse Tumbleweed system.


gdamore:
getrandom() may be a system call on your system, but that's not universal.  On others it may be implemented as a function.  This change is not acceptable as it isn't portable.

Question is why is getrandom() being detected but not actually present on your system?

gdamore:
Note that we use CMake's check_function_exists to detect the presence of getrandom().  Maybe you need to link against libc properly?

PR #1384, opened 2020-12-27
Pipeline

gdamore:

contributing guidelines.


gdamore:
This PR should be able to demonstrate a significant perf. boost for the pipe line consumers, while providing a more immediate feedback for work distribution.

gdamore:
Note that buffering is only available on the send side.  We do no receive side buffering.

gdamore:
My initial testing on Windows 10 (an Intel NUC that I use as my desktop, nothing special) shows 106K msgs/sec, and since path latency around 9us.

Note that in WSL the numbers fall *far* short of that -- about 9.2k msgs/sec, and around 95 us latency.  I'm going to blame WSL, but it's possible that there are also differences in the Linux implementation of locks and threading that make this so different.

Unfortunately I don't have a bare metal Linux box to test with.

PR #1402, opened 2021-01-03
fixes #1401 valgrind reports leaks in all tests

### LINKED ISSUE #1401: valgrind reports leaks in all tests
In pretty much every test, valgrind reports false leaks due to globals that we allocate.  We should arrange to call nng_fini via atexit in the test version of the library.  We can do this from nni_init actually.

gdamore:
This arranges for nng_fini to be called via atexit in the test
version of the library.

contributing guidelines.


src/core/init.c
```diff
@@ -30,7 +30,11 @@ nni_init_helper(void)
 	NNI_LIST_INIT(&nni_init_list, nni_initializer, i_node);
 	nni_inited = true;
 #ifdef NNG_TEST_LIB
-	atexit(nni_fini);
+	static bool cleanup = false;
+	if (!cleanup) {
```
unspecd, line 34: This check is redundant, isn't it?

in `posix/posix_thread.c`:
```c
int
nni_plat_init(int (*helper)(void))
{
	int rv;

	if (nni_plat_forked) {
		nni_panic("nng is not fork-reentrant safe");
	}
	if (nni_plat_inited) {
		return (0); // fast path
	}
...
```

gdamore: Well, the posix stuff is there, but I also want to call atexit only once.  I'm trying to find an elegant way to ensure that we call nng_fini() once at the end.   Unfortunately, I think this isn't working quite right yet.

unspecd: > Well, the posix stuff is there, but I also want to call atexit only once

But the `nni_init_helper` function should be called only once anyway:
```c
// The function argument should be called if the platform has not initialized
// (i.e. exactly once), and its result passed back to the caller.  If it
// does not return 0 (success), then it may be called again to try to
// initialize the platform again at a later date.
extern int nni_plat_init(int (*)(void));
```
`nni_init` -> `nni_plat_init` -> `nni_init_helper` (once) -> `atexit` -> `nni_fini`.

Or maybe I misunderstood something.

gdamore: Well, yes, unless someone calls nng_fini followed by nng_init.

I really really despise nng_fini -- its needed for valgrind and also to clean up allocations before unloading a shared library (in case someone uses this via dlopen().  Which I do not recommend!)

unspecd: Ah, got it.

PR #1423, opened 2021-02-04
Add --pipe option to nngcat

JWalker1995:
It would be nice for nngcat to be able to send a stream of data from a unix pipe (stdin or a named pipe file). This is impossible to do with the existing --file option, which reads the data into memory, then sends it one or more times. With the proposed --pipe option, each line from the specified pipe is sent as a message. This is more similar to how the `nc` and `socat` utilities function.

### DIFF
--- a/src/tools/nngcat/nngcat.c
+++ b/src/tools/nngcat/nngcat.c
@@ -44,6 +44,7 @@ nng_duration sendtimeo = NNG_DURATION_INFINITE;
 nng_duration recvtimeo = NNG_DURATION_INFINITE;
 void *       data      = NULL;
 size_t       datalen   = 0;
+FILE *       pipefile  = 0;
 int          compat    = 0;
 int          async     = 0;
 int          insecure  = 0;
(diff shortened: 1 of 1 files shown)


src/tools/nngcat/nngcat.c
```diff
@@ -705,6 +709,36 @@ sendrecv(nng_socket sock)
...
+	if (pipefile == NULL) {
+		fatal("No pipe to send (specify with --pipe)");
+	}
+	if (delay > 0) {
+		nng_msleep(delay);
+	}
+
+	int          rv;
+	nng_msg *    msg;
+  char * line = NULL;
```
gdamore, line 724: Minor indenting error.  We use indents by tab, where the tab is 8 cells wide.  Please see our .clang-format.

gdamore:
This PR is unfortunately port portable (see the windows failures).

May consider converting it to use stdio (which is supported under all ANSI C platforms, including Windows.)

dch:
this is a very nice feature - thanks!

gdamore (changes requested):
This needs to be converted to use fgets() or something portable in lieu of getline().

And I'd appreciate it if you could run this through clang-format.

wishdev:
I took the original commit, switched it to fgets and ran clang-format on the file on my branch here [1] - I'm never sure the best way to move this type of thing forward. Should I just create a new pull request?

[1] - https://github.com/wishdev/nng/tree/fgets

PR #1424, opened 2021-02-14
Remove extraneous/out-dated comment, declare conn_init static

voutilad:
Was looking into how to extend TLS support to a new engine when I found a small oddity.

The comment is an artifact from an old `tls_reap` function and no longer describes the following function (`conn_init`).

Function `conn_init` should be marked `static` as well to fit the design of the tls engine api and match the other mbed-related functions local to `src/supplemental/tls/mbedtls/tls.c`.

### DIFF
--- a/src/supplemental/tls/mbedtls/tls.c
+++ b/src/supplemental/tls/mbedtls/tls.c
@@ -167,9 +167,7 @@ conn_fini(nng_tls_engine_conn *ec)
 	mbedtls_ssl_free(&ec->ctx);
 }
 
-// The common code should call this only after it has released
-// it's upper layer stuff.
-int
+static int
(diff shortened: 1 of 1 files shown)


gdamore:
Which implementation were you trying to support ?

Get Outlook<https://aka.ms/qtex0l> for iOS
________________________________

voutilad:
I'm experimenting with `libtls` from the LibreSSL project, mostly since I've used it a few times and wanted to try wiring up a new engine to better understand nng's TLS workings. Given I'm working natively on OpenBSD a lot of the time, it's something easy for me to reach for that's familiar.

I'm skeptical it would be useful other than for my own education since mbedTLS is pretty universally available.

PR #1441, opened 2021-04-08
make getrandom a syscall, so it doesn't cause missing symbol errors

marijnfs:
This calls getrandom as a syscall.
'getrandom' seems to be a syscall, not just a function you can call. This causes troubles in certain settings (depends on your standard library i guess) during linking ('getrandom' will be a missing symbol).

### DIFF
--- a/src/platform/posix/posix_rand_getrandom.c
+++ b/src/platform/posix/posix_rand_getrandom.c
@@ -15,7 +15,7 @@
 // normally be running.  This is the only time it can fail with correct
 // arguments, and then only if it is interrupted with a signal.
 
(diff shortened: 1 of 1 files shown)


gdamore:
The approach provided is not portable, and thus unacceptable.

If you're running into this problem, its likely because the library was built on a system with a standard library that has this function, but you are running it on another system where the standard library does not.  This is a problem in the way your artifacts are constructed.

I'm closing this -- if you want to come back with a different approach to improve the detection, that would be fine, but I'm very much opposed to introducing incredibly platform specific unless there is a compelling reason.  I am not seeing it here.

PR #1449, opened 2021-04-22
Turn aio expire queue from nni list to array for efficiency.

wanghaEMQ:
When there are too many aios in the nni list, the efficiency of the ordered list decreases significantly.
Turn aio expire queue from ordered nni list to array will perform better.

The perf results in projects nanomq(forked from nng) are presented in following two xml file.
Test conditions.

+ 25k connections.
+ interval of connection is 6 seconds.
+ each test lasts about 3 minutes.

(Notes. xml format is not supported for attachments. so, i compress it with tar.)
[perf.tar.gz](https://github.com/nanomsg/nng/files/6357856/perf.tar.gz)

### DIFF
--- a/src/core/aio.c
+++ b/src/core/aio.c
@@ -12,12 +12,14 @@
 #include <string.h>
 
 struct nni_aio_expire_q {
-	nni_mtx  eq_mtx;
-	nni_cv   eq_cv;
-	nni_list eq_list;
-	nni_aio *eq_aio; // currently expiring (task dispatch)
-	nni_thr  eq_thr;
-	bool     eq_exit;
+	nni_mtx   eq_mtx;
+	nni_cv    eq_cv;
+	nni_aio **eq_list;
+	uint32_t  eq_len;
+	uint32_t  eq_cap;
+	nni_aio  *eq_aio; // currently expiring (task dispatch)
+	nni_thr   eq_thr;
+	bool      eq_exit;
 };
 
 static nni_aio_expire_q **nni_aio_expire_q_list;
@@ -98,6 +100,7 @@ static nni_reap_list aio_reap_list = {
 };
 
 static void nni_aio_expire_add(nni_aio *);
+static void nni_aio_expire_rm(nni_aio *);
 
 void
 nni_aio_init(nni_aio *aio, nni_cb cb, void *arg)
@@ -123,7 +126,7 @@ nni_aio_fini(nni_aio *aio)
 	// We also wait if the aio is being expired.
 	nni_mtx_lock(&eq->eq_mtx);
 	aio->a_stop = true;
-	nni_list_node_remove(&aio->a_expire_node);
+	nni_aio_expire_rm(aio);
 	while (eq->eq_aio == aio) {
 		nni_cv_wait(&eq->eq_cv);
 	}
@@ -204,7 +207,7 @@ nni_aio_stop(nni_aio *aio)
 		nni_aio_expire_q *eq = aio->a_expire_q;
 
 		nni_mtx_lock(&eq->eq_mtx);
-		nni_list_node_remove(&aio->a_expire_node);
+		nni_aio_expire_rm(aio);
 		fn                = aio->a_cancel_fn;
 		arg               = aio->a_cancel_arg;
 		aio->a_cancel_fn  = NULL;
@@ -229,7 +232,7 @@ nni_aio_close(nni_aio *aio)
 		nni_aio_expire_q *eq = aio->a_expire_q;
 
 		nni_mtx_lock(&eq->eq_mtx);
-		nni_list_node_remove(&aio->a_expire_node);
+		nni_aio_expire_rm(aio);
 		fn                = aio->a_cancel_fn;
 		arg               = aio->a_cancel_arg;
 		aio->a_cancel_fn  = NULL;
@@ -409,7 +412,7 @@ nni_aio_abort(nni_aio *aio, int rv)
 	nni_aio_expire_q *eq = aio->a_expire_q;
 
 	nni_mtx_lock(&eq->eq_mtx);
-	nni_list_node_remove(&aio->a_expire_node);
+	nni_aio_expire_rm(aio);
 	fn                = aio->a_cancel_fn;
 	arg               = aio->a_cancel_arg;
 	aio->a_cancel_fn  = NULL;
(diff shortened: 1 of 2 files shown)


src/core/aio.c
```diff
@@ -505,6 +508,7 @@ nni_aio_list_active(nni_aio *aio)
 static void
 nni_aio_expire_add(nni_aio *aio)
 {
+	/*
```
gdamore, line 511: Please don't just comment out code -- remove it.   It makes it easier to review what you're changing, and keeps us from accruing cruft.

src/core/aio.c
```diff
@@ -523,16 +527,61 @@ nni_aio_expire_add(nni_aio *aio)
 		// And, as we are the latest, kick the thing.
 		nni_cv_wake(&aio->a_expire_q->eq_cv);
 	}
+	*/
+
+	nni_aio_expire_q * eq = aio->a_expire_q;
+	if (eq->eq_len > eq->eq_cap * 2 / 3) {
```
gdamore, line 533: growing at 2/3?  why not just wait until you actually grow exactly to the point of needing it?

gdamore: also this division by three is both expensive and unnecessary.

src/core/aio.c
```diff
@@ -523,16 +527,61 @@ nni_aio_expire_add(nni_aio *aio)
...
+	for (aio_idx = 0; aio_idx < eq->eq_len; aio_idx++) {
+		if (aio == eq->eq_list[aio_idx]) {
+			for (uint32_t i = aio_idx + 1; i < eq->eq_len; i++) {
+				eq->eq_list[i - 1] = eq->eq_list[i];
+			}
+			eq->eq_len --;
+			break;
+		}
+	}
+
```
gdamore, line 563: i'd be reasonably inclined *not* to automatically shrink the list... or to have the size at which we shrink be a tunable.  I would not ever expect the size of this array to be a serious problem -- for example if you have 1M items outstanding for expiration, you're only going to need 8MB.  That's an extreme number of aios (outside the real world for certain!) while still only needing a relatively modest queue size.

Avoiding the churn on the allocator is a very worthwhile thing to do.

src/core/aio.c
```diff
@@ -556,16 +605,29 @@ nni_aio_expire_loop(void *arg)
...
+		aio = list[aio_idx];
+		for (uint32_t i = 0; i < q->eq_len; i++) {
+			if (list[i]->a_expire < aio->a_expire) {
+				aio = list[i];
+				aio_idx = i;
+			}
+		}
 		if (now < aio->a_expire) {
-			// Unexpired; the list is ordered, so we just wait.
+			// Unexpired; we just wait the next expired aio.
```
gdamore, line 619: we just wait *for* the next...

src/core/aio.c
```diff
@@ -556,16 +605,29 @@ nni_aio_expire_loop(void *arg)
...
 			nni_cv_until(cv, aio->a_expire);
 			now = nni_clock();
 			continue;
 		}
 
 		// The time has come for this aio.  Expire it, canceling any
 		// outstanding I/O.
-		nni_list_remove(list, aio);
+		for (uint32_t i = aio_idx+1; i < q->eq_len; i++) {
```
gdamore, line 627: this is silly.  the list is not ordered, so you don't need to do copy them all down.

you can just move the last element here instead.  this turns O(n) to just a single operation.

gdamore (changes requested):
Conceptually I like what is here -- it is what we discussed with Jaylin after all.

However I see some easy ways to make this even simpler and more efficient.

Thank you for adhering to the style as well.

src/core/aio.c
gdamore: btw, this can be done without a condition -- if this was the last item, then the swap will just move the item to itself, which won't matter, because you should then decrement the size.  (important: do the swap before you change the size.)  avoiding conditions is good because it means fewer pipeline stalls in the cpu.

src/core/aio.c
JaylinYu: Hi Garrett, Totally agree with you. Hence, we never shrink the list or leave a tunable CMAKE parameter like NNG_MAX_AIO_LIST to users so that they can have control over the exact memory is consumed?

src/core/aio.c
```diff
@@ -505,34 +508,57 @@ nni_aio_list_active(nni_aio *aio)
 static void
 nni_aio_expire_add(nni_aio *aio)
 {
-	nni_list *list = &aio->a_expire_q->eq_list;
-	nni_aio * prev;
+	nni_aio_expire_q * eq = aio->a_expire_q;
+	if (eq->eq_len > eq->eq_cap) {
+		nni_aio ** new_list =
```
gdamore, line 513: very minor code formatting nit.  Please run this code through clang format.

We use the unary pointer operator "*" without a space between the variable.

For example:

char *ptr

instead of 

char * ptr

Thanks.

src/core/aio.c
```diff
@@ -505,34 +508,57 @@ nni_aio_list_active(nni_aio *aio)
...
-			nni_list_insert_after(list, aio, prev);
+static void
+nni_aio_expire_rm(nni_aio *aio)
+{
+	uint32_t          aio_idx = 0;
+	nni_aio_expire_q *eq      = aio->a_expire_q;
+	for (aio_idx = 0; aio_idx < eq->eq_len; aio_idx++) {
+		if (aio == eq->eq_list[aio_idx]) {
+			eq->eq_list[aio_idx] = eq->eq_list[eq->eq_len-1];
+			eq->eq_len --;
```
gdamore, line 536: yeah another one that clang format should fix.  eq->eq_len-- ;   (no space before the --)

src/core/aio.c
```diff
@@ -505,34 +508,57 @@ nni_aio_list_active(nni_aio *aio)
...
 			break;
 		}
 	}
-	if (prev == NULL) {
-		// This has the shortest time, so insert at the start.
-		nni_list_prepend(list, aio);
-		// And, as we are the latest, kick the thing.
-		nni_cv_wake(&aio->a_expire_q->eq_cv);
+
+	if (eq->eq_len < eq->eq_cap / 3 && eq->eq_cap > NNI_EXPIRE_Q_SIZE) {
```
gdamore, line 541: Please consider changing / 3 to / 4.  Division by 3 is very expensive.  Division by 4 (or any other power of 2) is super cheap.  This might make it slower to reclaim memory, but really, that's probably ok.  If we bursted up to this value, we might burst again, and using a quarter as the low threshold to free should further reduce pressure on the allocator for bursty workloads.

gdamore (approved):

gdamore:
I merged this -- the UBsan problems are not yours.

PR #1458, opened 2021-07-06
Add & document msg_capacity, msg_reserve

EvanBalster:
This PR creates and documents two new functions dealing with message capacity — the maximum length to which a message can grow without invoking the memory allocator.

`nng_msg_capacity` returns the current capacity based on the underlying body chunk (not counting headroom)
`nng_msg_reserve` will expand capacity to be greater than or equal to the given value, but will in no case decrease message length or headroom.  (Under the hood, it forwards to `nni_chunk_grow`)

Please double check that my changes to the documentation were done correctly.


src/core/message.c
```diff
@@ -456,6 +456,18 @@ nni_msg_realloc(nni_msg *m, size_t sz)
 	return (0);
 }
 
+int
+nni_msg_reserve(nni_msg *m, size_t capacity)
+{
+	return nni_chunk_grow(&m->m_body, capacity, 0);
```
gdamore, line 462: style nit:

we parenthesize the return value always.

src/core/message.c
```diff
@@ -456,6 +456,18 @@ nni_msg_realloc(nni_msg *m, size_t sz)
...
+}
+
+size_t
+nni_msg_capacity(nni_msg *m)
+{
+	return (size_t) ((m->m_body.ch_buf + m->m_body.ch_cap) - m->m_body.ch_ptr);
```
gdamore, line 468: ditto the style -- in this case the parenthesis should include the cast.

e.g.

return ((size_t) (....))

docs/man/nng_msg_capacity.3.adoc
```diff
@@ -0,0 +1,46 @@
+= nng_msg_capacity(3)
```
gdamore, line 1: Can you add a few more cross references back to these pages?

Specifically in nng_msg.5.adoc, and nng_msg_len.3.adoc and nng_msg_realloc.3.adoc?

gdamore:
Looks pretty good, except for a few more doc items and a style nits.

neachdainn:
Do you want to always grow the chunk with `nng_msg_reserve`? It might make sense to add a check against the current capacity.

docs/man/nng_msg_capacity.3.adoc
EvanBalster: There's already a cross-reference in realloc (possibly I missed the list at the bottom, though).  I'll attend to the others.

gdamore:
> Do you want to always grow the chunk with `nng_msg_reserve`? It might make sense to add a check against the current capacity.

This isn't necessary -- nni_chunk_grow already has these checks and won't actually change the message if the capacity is already as large as (or larger than) the requested capacity.

EvanBalster:
Style nits addressed, cross-references added.

> Do you want to always grow the chunk with `nng_msg_reserve`? It might make sense to add a check against the current capacity.

I think this is the behavior of `nni_chunk_grow`.

gdamore:
Prolly we should have put some test case coverage on this.  However, tests/message.c really needs to be converted over to NUTS so I don't particularly feel like asking you to fix that.

PR #1469, opened 2021-07-13
Fix valgrind error epoll_ctl points to uninitialised bytes

Wramberg:
I am trying to replace ZeroMQ with NNG in some company libraries. In our UTs I have added `nng_fini()` to our `atexit()` to avoid false positives reported by valgrind regarding possible lost bytes. One thing I cannot get rid of however is this error

```
==1664== Syscall param epoll_ctl(event) points to uninitialised byte(s)
==1664==    at 0x4D527BE: epoll_ctl (syscall-template.S:78)
==1664==    by 0x49660D8: nni_posix_pfd_arm (in /usr/lib/i386-linux-gnu/libnng.so.1)
==1664==    by 0x495FACF: ipc_listener_doaccept (in /usr/lib/i386-linux-gnu/libnng.so.1)
==1664==    by 0x4960799: ipc_listener_accept (in /usr/lib/i386-linux-gnu/libnng.so.1)
==1664==    by 0x49584EA: nng_stream_listener_accept (in /usr/lib/i386-linux-gnu/libnng.so.1)
==1664==    by 0x497B3FF: ipc_ep_accept (in /usr/lib/i386-linux-gnu/libnng.so.1)
==1664==    by 0x494F16E: listener_accept_start (in /usr/lib/i386-linux-gnu/libnng.so.1)
==1664==    by 0x494F1F4: nni_listener_start (in /usr/lib/i386-linux-gnu/libnng.so.1)
==1664==    by 0x494656D: nng_listen (in /usr/lib/i386-linux-gnu/libnng.so.1)
```

and another, similar one (this PR targets both). I realize that this is a good candidate for a valgrind suppression but I wanted to see where NNG stands on this issue. For me, using suppressions would mean having a suppression file in _many_ unittests as the libraries are used in many places.

Alternatively, I have on some occasions simply changed my code to work better with the tool in question. My intention with this PR is to discuss whether this could be considered for NNG ? Note the changes proposed here are merely to illustrate what I am talking about in the codebase.

### DIFF
--- a/src/platform/posix/posix_pollq_epoll.c
+++ b/src/platform/posix/posix_pollq_epoll.c
@@ -107,6 +107,7 @@ nni_posix_pfd_init(nni_posix_pfd **pfdp, int fd)
 	NNI_LIST_NODE_INIT(&pfd->node);
 
 	// notifications disabled to begin with
+	memset(&ev, 0, sizeof(ev));
 	ev.events   = 0;
 	ev.data.ptr = pfd;
 
@@ -138,6 +139,7 @@ nni_posix_pfd_arm(nni_posix_pfd *pfd, unsigned events)
 		pfd->events |= events;
 		events = pfd->events;
 
+		memset(&ev, 0, sizeof(ev));
 		ev.events   = events | NNI_EPOLL_FLAGS;
 		ev.data.ptr = pfd;
 


gdamore:
The data structures in this case should probably have been allocated with nni_zalloc() -- I try to avoid explicit calls to memset or bzero.   The reason I did not do this was for performance -- if we don't care about the bytes/fields, then we don't need to waste cycles initializing them.

Still if this is one of the things that causes false complaints -- it might be worth doing the zalloc for those reasons.  I'd rather have the code use nni_zalloc at allocation time if possible.

Wramberg:
Completely understandable starting point. Yeah I imagined my rudimentary approach wasn't suited for merging :) Well I can only speak for myself but I am getting false complaints from `valgrind-3.13.0`. 

Ill change it to use `zalloc` and verify again.

Right now I am using a suppression file to keep valgrind happy but as stated I would prefer to avoid having to bring this with me everywhere I update to use NNG

```
{
   nng-epoll-false-positive
   Memcheck:Param
   epoll_ctl(event)
   fun:epoll_ctl
   fun:nni_posix_pfd_init
}
{
   nng-epoll-false-positive2
   Memcheck:Param
   epoll_ctl(event)
   fun:epoll_ctl
   fun:nni_posix_pfd_arm
}
```

src/platform/posix/posix_pollq_epoll.c
```diff
@@ -107,17 +112,19 @@ nni_posix_pfd_init(nni_posix_pfd **pfdp, int fd)
 	NNI_LIST_NODE_INIT(&pfd->node);
 
 	// notifications disabled to begin with
-	ev.events   = 0;
-	ev.data.ptr = pfd;
+	ev->events   = 0;
+	ev->data.ptr = pfd;
 
-	if (epoll_ctl(pq->epfd, EPOLL_CTL_ADD, fd, &ev) != 0) {
+	if (epoll_ctl(pq->epfd, EPOLL_CTL_ADD, fd, ev) != 0) {
```
Wramberg, line 118: Could consider reading the result into a variable and free'ing `ev` right after before we start checking return codes ?

src/platform/posix/posix_pollq_epoll.c
```diff
@@ -134,19 +141,27 @@ nni_posix_pfd_arm(nni_posix_pfd *pfd, unsigned events)
...
 	if (!pfd->closing) {
-		struct epoll_event ev;
+		struct epoll_event *ev;
 		pfd->events |= events;
 		events = pfd->events;
 
-		ev.events   = events | NNI_EPOLL_FLAGS;
-		ev.data.ptr = pfd;
+		if ((ev = NNI_ALLOC_STRUCT(ev)) == NULL) {
+			return (NNG_ENOMEM);
```
Wramberg, line 149: From what I can tell it is ok to introduce the `NNG_ENOMEM` return code here. Seems the top-level functions already state that `NNG_ENOMEM` can be returned from them. Not entirely confident in my quick glance over the source code though.

src/platform/posix/posix_pollq_epoll.c
```diff
@@ -134,19 +141,27 @@ nni_posix_pfd_arm(nni_posix_pfd *pfd, unsigned events)
...
+		}
+
+		ev->events   = events | NNI_EPOLL_FLAGS;
+		ev->data.ptr = pfd;
 
-		if (epoll_ctl(pq->epfd, EPOLL_CTL_MOD, pfd->fd, &ev) != 0) {
+		if (epoll_ctl(pq->epfd, EPOLL_CTL_MOD, pfd->fd, ev) != 0) {
```
Wramberg, line 155: Same as above. Could consider reading return value and freeing before we check it.

Wramberg:
@gdamore Updated to `nni_zalloc` as discussed. Thoughts ?

src/platform/posix/posix_pollq_epoll.c
gdamore: Actually looking at this again, I realize I was using a variable on the stack.  I didn't notice that earlier.

This was to avoid extra code paths (you have to handle NNG_ENOMEM) as well as reduce pressure on the allocator (allocations are often expensive). 

I am now thinking that valgrind's analysis is wrong here.  It's casting too wide a net -- having a variable that points at uninitialized memory is fine, as long as I don't do any reads -- valgrind used to be smarter about this.

These days I tend to prefer to the CLang analyzers that I think do a *vastly* better job than valgrind -- they catch more issues than valgrind, run faster, and don't give me false positives.

At this point, I think I'm disinclined to make this change -- "fixing" it is really working around an incorrect assessment in valgrind, and it costs something to do so.

gdamore: Looking again, its a complaint that the syscall parameters are uninitialized, and valgrind can't reach across the syscall boundary.

Maybe the memset you originally proposed is best.

Wramberg:
> 
> Actually looking at this again, I realize I was using a variable on the stack. I didn't notice that earlier.

Must admit the change to zalloc felt weird given that `ev` was originally on the stack. 

> 
> I am now thinking that valgrind's analysis is wrong here. It's casting too wide a net -- having a variable that points at uninitialized memory is fine, as long as I don't do any reads -- valgrind used to be smarter about this.

Yea I have no doubt it is not causing any issues. Hence my original comment about making a suppression for it.

> 
> These days I tend to prefer to the CLang analyzers that I think do a _vastly_ better job than valgrind -- they catch more issues than valgrind, run faster, and don't give me false positives.
> 

Generally I prefer those as well but we our existing infrastructure include valgrind so Ill have to accommodate that.

> At this point, I think I'm disinclined to make this change -- "fixing" it is really working around an incorrect assessment in valgrind, and it costs something to do so.

This was my key question here. To figure out where you stand on this topic and whether I have to drag suppressions around or if we could work around valgrind false positives.


> 
> Looking again, its a complaint that the syscall parameters are uninitialized, and valgrind can't reach across the syscall boundary.
> 
> Maybe the memset you originally proposed is best.

I can revert back to `memset` no problem - but if you are disinclined to make the change anyway I guess it doesn't matter ? I would prefer to make this change but I can understand if you decline given that it does come with a slight performance cost.

Wramberg:
@gdamore Any thoughts on this ?

> I can revert back to memset no problem - but if you are disinclined to make the change anyway I guess it doesn't matter ? I would prefer to make this change but I can understand if you decline given that it does come with a slight performance cost.

gdamore:
Let's do the memset thing.  I agree that it's a bad idea to push down uninitialized buffers to system calls - particularly if the entire buffer will not be overwritten (as in read). The performance cost is minimal enough that I'm willing to go that route.

Wramberg:
@gdamore Done :)

PR #1472, opened 2021-07-18
refactor(docs): refactor man & index page

oceanlvr:
The home page and man page have been redesigned.


oceanlvr:
@gdamore PTAL :)

assets/css/nn.css
```diff
@@ -118,63 +50,43 @@ body {
...
-        font-size: 1.6875em;
-}
-
-h4 {
-        font-size: 1.4375em;
-}
-
-h5 {
-        font-size: 1.2em;
+        font-size: 16px;
```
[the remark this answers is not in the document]
oceanlvr: TL;DR; Our website is mainly for `PC`, use adaptive `font-size` base on viewport layout in `PC` is strange, and the fixed size on mobile is just bigger. (acceptable, our customers also will focus on the font, not the picture., and so sorry we don't have a mobile design draft :(  ).

> em = desired element pixel value / parent element font-size in pixels

`em` is a computing unit, which bases its `father`. it won't be adaptive normally unless we assign his father an adaptive value. and sometimes we could not be guaranteed the size of his father's `font-size`. 

so `em` is not a good adaptive unit.

```
html {
  font-size: 16px;
}

h1 {
  font-size: 2em; // dangerous, its father is not sure, maybe not html
}
```

-------------

`rem` units are pretty like `em`, but it bases on the root element's `font-size`. and `vh/vw` base on viewport.

we could combine them and make the website adaptive.

```
html {
  font-size: 26.67vw; // base on viewport(26.67% of the viewport width), adaptive unit
  margin: 0 auto;
}
body {
  font-size: 0.14rem; // equal 26.67vw * 0.14 rem, compute unit
}
```

BUT our website mainly bases on `PC`.

For `PC`, when we magnify the viewport, the font size magnifies too. but usually, we don't like font-size changed. (`font-size` on `1024` viewport, width is not the same as `2048`, which is strange).

the best practice I think is using media queries for mobile device, and combine `px` (for PC) and `vh/vw` with `rem` (for mobile)

```css
html {
  font-size: 16px;
}

h1, h2 {
  font-size: 16px;
}

@media  screen and (max-width: 1024px) {
  html {
    font-size: 26.67vw;
  }
  body {
    font-size: 0.14rem; 
  }
}
```

and we don't design this part. so just use fixed is OK. Our customers also will focus on the font, not the picture.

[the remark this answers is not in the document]
oceanlvr: `em` could inherit its father(if his father assign or just inherit from browser).

we should not use `em` with `px`. it's dangerous.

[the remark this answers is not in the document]
gdamore: I was thinking em, but where you need stricter control (and don't want the size to change based on the context it appears in) rem is better.

I'm concerned that by fixing the pixels, the design is "physical" and may not scale well on all displays, and may also not honor user preferences.

While most folks coming to this site are likely coming from a regular computer, some effort was invested to make sure that the site would adapt to other conditions before, which is why the old font sizes were in em.

I guess however that these days many other sites are using physical markup in px instead of em/rem, and so folks who need to adjust have probably already done so.  I'm going to take a look again.

_includes/header-nng.html
```diff
@@ -1,20 +1,71 @@
...
+                                                <a href="/man/">Documentation</a>
+                                        </div>
+                                        <div class="navbar-item">
+                                                <a href="https://discord.gg/Xnac6b9">Discord</a>
+                                        </div>
+                                        <div class="navbar-item is-hidden-desktop">
+                                                <a href="https://github.com/nanomsg/nng">
+                                                        github
+                                                </a>
+                                        </div>
```
gdamore, line 28: Probably it should be GitHub properly capitalized.

_includes/header-nng.html
```diff
@@ -1,20 +1,71 @@
...
+                                        <div class="navbar-item is-hidden-touch">
+                                                <a href="https://github.com/nanomsg/nng" class="btn-github">
+                                                        <i class="iconfont icon-github"></i> Star
```
gdamore, line 31: I think this should say GitHub instead of Star.   The button isn't letting you change stars, and I would not really like it to either.

_includes/header-nng.html
```diff
@@ -1,20 +1,71 @@
...
+                                </a>
+                        </div>
+                        <div id="navbarBasicExample" class="navbar-menu">
+                                <div class="navbar-start"></div>
+                                <div class="navbar-end">
+                                        <div class="navbar-item">
+                                                <a href="/man/">Documentation</a>
+                                        </div>
+                                        <div class="navbar-item">
+                                                <a href="https://discord.gg/Xnac6b9">Discord</a>
```
gdamore, line 22: Given you have a GitHub icon, maybe also a Discord icon (and maybe a Docs icon of some kind?)

_includes/nn-content.html
```diff
@@ -0,0 +1,158 @@
...
+  <section id="nn-quick-start">
+    <div class="container">
+      <h2 class="nn-subtitle has-text-centered nn-quick-start-title has-text-white">
+        Quick Start
+      </h2>
+      <div class="columns is-8-desktop is-variable">
+        <div class="column is-6">
+          <!-- TODO: add a labels' link -->
+          <p class="nn-quick-start-desc mb-3">
+            To build this project, you will need a C99 compatible compiler and <a class="is-brown" href="">CMake</a> version 3.13 or
```
gdamore, line 133: The brown color link, on the slate purple or whatever color it is, has pretty poor contrast making it harder to read.  We should pick a color that contrasts better.  Maybe even black, or a more intense form of brown?

_includes/header-nng.html
oceanlvr: Yes, it might be better use `Github`

_includes/nn-content.html
oceanlvr: I agree that I may need to discuss it with my colleagues. Brown is really not very friendly.

assets/css/nn.css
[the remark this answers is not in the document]
oceanlvr: Yes. Using `em` may be more friendly to some users who have a preference for font control. 

For web pages that are primarily for the `PC` side, we take control for the sake of layout certainty which ensures that the layout is consistent and simple. :)

oceanlvr:
@gdamore Please takes a look.

Issue #1486, opened 2021-08-09
nng_sleep_aio never wake up while hitting NNG_ENOFILES


JaylinYu:
**Describe the bug**
Whole nng system hangs and aio_expire_q_loop got stucked (nng_sleep_aio(10, ep->timeaio); never expired ) while OS runs out of FD (hitting emfiles) and the cb function (tcptran_accept_cb) of epoll/poll gets NNG_ENOFILES.

**Expected behavior**
nng should keep wake up threads after back off certain time.

**Actual Behavior**
aio never finished after nng_sleep_aio.

**To Reproduce**
set ulimt -n 256
and connect to any of nng port with more than 256 clients.

** Environment Details **
 - NNG version 1.5.1
 - Operating system and version Linux

**Additional context**
there is a temporary fix coming soon.

gdamore:
This was a regression caused by the aio expiration work introduced in #1449 - f37a72f3697125092baa510f26d2e2f9f0b854e0

This affects all 1.5.x releases, and it isn't required to run out of files to hit the problem -- basically all it requires is the expiration ordering not being correct.

This particular fix is band-aid that resolves the problem, but as per discussion with the contributor, there is further work to be done to provide the main performance benefits intended by (but not achieved by) #1449.

PR #1490, opened 2021-08-10
fixes #1488 aio expiration list performance work needed

### LINKED ISSUE #1488: aio expiration list performance work needed
The fix for #1449 was unfortunately not quite complete, and introduced a regression. See #1486.

While the regression is now fixed, we are leaving a lot of performance on the table, and have regressed back to O(n^2) on the expiration list.

The conversion to an array should be changed to retain the index so that during removal we can recover the slot without having to look for it.  As this is the hot code path, this will save a lot of cycles.

Additionally we need to keep the "next expiration time" in a per expiration queue variable, so that we can avoid waking up the expiration loop unless that loop needs to be woken for rescheduling because the newly inserted item expires sooner than any other item on the array (or is the only item on the array).

gdamore:
There were several problems with the array implementation, both
from performance and from correctness.  This corrects those errors
(hopefully) and restores the expiration lists as linked lists.

### DIFF
--- a/src/core/aio.c
+++ b/src/core/aio.c
@@ -12,14 +12,13 @@
 #include <string.h>
 
 struct nni_aio_expire_q {
-	nni_mtx   eq_mtx;
-	nni_cv    eq_cv;
-	nni_aio **eq_list;
-	uint32_t  eq_len;
-	uint32_t  eq_cap;
-	nni_aio * eq_aio; // currently expiring (task dispatch)
-	nni_thr   eq_thr;
-	bool      eq_exit;
+	nni_mtx  eq_mtx;
+	nni_cv   eq_cv;
+	nni_list eq_list;
+	uint32_t eq_len;
+	nni_thr  eq_thr;
+	nni_time eq_next; // next expiration
+	bool     eq_exit;
 };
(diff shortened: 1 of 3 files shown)


JaylinYu:
Hi Garrett, I can confirm this PR does solve the bug #1486. However, as my roughly benchmarked, there is no much difference from the old code(nni_list) in terms of performance.

gdamore:
There was a pathological performance problem in the old code (O(n^2)). Are still seeing that performance cliff and is it in the same place?

The new code tries to make both adding and removing from the list in constant time. The expiration loop runs less frequently but when it does it has to perform a linear time scan. 

Another option would be to turn this into a heap but I think that that might not help that much.  We would be O(log(n)) for every aio instead of just linear for some of them.

Anyway I would very much like to see benchmarks. I'm skeptical that there is that much of a win in the array vs list since you have to dereference the aio anyway so there should not be any extra cache penalty. (Except.  Maybe the cost to dereference the *next* item on the list when adjusting the list. That happens when adding or removing but with the array it would only happen when iterating.)

At any rate some quantitative results would be helpful here.

PR #1511, opened 2021-09-24
Reduce SP code duplication

unspecd:
Reducing code duplication in `src/sp/transport`. Hope you will find this useful.

### DIFF
--- a/src/sp/transport/ipc/ipc.c
+++ b/src/sp/transport/ipc/ipc.c
@@ -21,835 +21,17 @@
 // Windows named pipes.  Other platforms could use other mechanisms,
 // but all implementations on the platform must use the same mechanism.
 
-typedef struct ipc_pipe ipc_pipe;
-typedef struct ipc_ep   ipc_ep;
-
-// ipc_pipe is one end of an IPC connection.
-struct ipc_pipe {
-	nng_stream *    conn;
-	uint16_t        peer;
-	uint16_t        proto;
-	size_t          rcv_max;
-	bool            closed;
-	nni_sockaddr    sa;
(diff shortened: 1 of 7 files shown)


unspecd:
(cleanup and unification)

src/sp/transport/ipc/ipc.c
```diff
@@ -21,835 +21,17 @@
...
-	}
-	if ((aio = nni_list_first(&p->send_q)) == NULL) {
-		return;
-	}
-
-	// This runs to send the message.
-	msg = nni_aio_get_msg(aio);
-	len = nni_msg_len(msg) + nni_msg_header_len(msg);
-
-	p->tx_head[0] = 1; // message type, 1.
```
gdamore, line 488: I think you may have overlooked this.  The wire protocol for IPC is slightly different than the other streaming protocols, because it adds another field (the message type).  If this change is not handled, it will no longer be compatible with other SP implementations for IPC.

gdamore (commented):
I'm also working on a substantial refactor to support different protocols (MQTT) coexisting, so I'm not sure I want to take this code yet -- I'm concerned it will add complexity to that effort.

src/sp/transport/ipc/ipc.c
unspecd: > I think you may have overlooked this.
> ...
> it will no longer be compatible with other SP implementations for IPC.

You are right, I got too carried away :)

unspecd:
> I'm also working on a substantial refactor to support different protocols

Okay, thanks for this great work. I will then close the PR.

PR #1515, opened 2021-10-07
Use env shebangs everywhere

leonardp:
Change all shebangs to use '#!/usr/bin/env bash'.
This increases portability to platforms which do not cohere to the FHS.

---
I am using NixOS as my development platform where "normal" shebangs do not work.
I don't know if using "env shebangs" lead to problems on other platforms, so feel free to close if this patch leads
to problems elsewhere.


gdamore:
If it passes workflow, I will approve.

PR #1525, opened 2021-11-16
Add static library compile option "-fPIC"

### LINKED ISSUE #1524: relocation against
**Describe the bug**
The issue happened while building third-party **dynamic** library with nng **static** library :

/usr/bin/ld: /usr/local/lib/libnng.a(inproc.c.o): warning: relocation against `nni_inproc_tran' in read-only section `.text'
/usr/bin/ld: /usr/local/lib/libnng.a(posix_debug.c.o): relocation R_X86_64_PC32 against symbol `stderr@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC

**Expected behavior**
Add compile option "-fPIC"

**Actual Behavior**
Describe what occurred.

**To Reproduce**
If possible include actual reproduction test code here.
Minimal C test cases are perferred.

** Environment Details **
 - NNG version: 1.5.2 
 - Operating system and version: Ubuntu 20.04
 - Compiler and language used: C
 - Shared or static library: static 

**Additional context**

alvin1221:
Fixes  #1525

> relocation against

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,6 +75,7 @@ add_definitions(-DNNG_PRIVATE)
 
 if (NOT (BUILD_SHARED_LIBS))
     set(NNG_STATIC_LIB ON)
+    add_compile_options(-fPIC)
     message(STATUS "Building static libs.")
 endif ()
 


gdamore:
When building a static library, it generally is *NOT* necessary to use position independent code.  The problem here stems when some project is mixing static and dynamic modules incorrectly.

If the intent is to build a version of NNG inside another module that is dynamic, without NNG itself being an external library, then the correct answer is to override the global flags in the consuming project.

It is not correct to make all projects use PIC.

PR #1531, opened 2021-11-24
update demos to not use deprecated functions

urkle:
Update the demos to not use deprecated functions


gdamore:
Thanks for this...

PR #1532, opened 2021-11-24
mark deprecated functions as deprecated so they show as warnings during compile

urkle:
defines a NNG_DEPRECATED macro that will emit the deprecated attribute supported by GCC & Clang
Add this macro to all deprecated functions

### DIFF
--- a/include/nng/nng.h
+++ b/include/nng/nng.h
@@ -45,6 +45,14 @@ extern "C" {
 #endif // _WIN32 && !NNG_STATIC_LIB
 #endif // NNG_DECL
 
+#ifndef NNG_DEPRECATED
+#if defined(__GNUC__) || defined(__clang__)
+#define NNG_DEPRECATED __attribute__ ((deprecated))
+#else
+#define NNG_DEPRECATED
+#endif
+#endif
+
 // NNG Library & API version.
 // We use SemVer, and these versions are about the API, and
 // may not necessarily match the ABI versions.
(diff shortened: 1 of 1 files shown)


gdamore:
What happens if people try to build with older versions of GCC or Clang?   Will these attribute definitions be ignored, or will the build fail?

urkle:
@gdamore you'd have to go to a REALLY REALLY REALLY old version of GCC to not have that supported.

Looking back through the GCC manuals those have been supported since GCC 3.2.x ( https://gcc.gnu.org/onlinedocs/gcc-3.2.3/gcc/Function-Attributes.html#Function%20Attributes ) which support C99 as well.  And GCC version was released in 2002!

I can put an additional guard in there is you want.

gdamore:
Ok, I'm not too worried about GCC 2.x or something like that.  But if there is a good way to test for this support without an explicit version check that might be nice.  These attribute things are part of newer standards, so perhaps we can use some standard-defined test macro?

urkle:
>  But if there is a good way to test for this support without an explicit version check that might be nice. 

I can see if I can do some CMAKE-foo to test if they are supported.

> These attribute things are part of newer standards, so perhaps we can use some standard-defined test macro?

So what I implemented is the "old" attributes extensions in GNUC/clang. (there is one for MSVC I can throw in too).
The "new" standards using `[[deprecated("reason")]]` are for C++11  or in C23 (not released yet). So no "clean" way of testing them there.

gdamore:
Fair enough.  I was hoping that support for this attribute was testable by a macro.  If it isn't then this is the best we can do.

gdamore:
Can you resolve the conflicts.  I think I'm ready to merge once that's done.

urkle:
@gdamore rebased and conflicts resolved for you

PR #1549, opened 2021-12-08
Fixes compiling on windows 32-bit

urkle:
- Use correct datatype for one nni_atomic_flag (that hadn't been changed to LONG)
   - resolves a compiler warning
- Use InterlockedExchangeAdd instead of InterlockedAdd (which isn't implemented on 32-bit windows)


gdamore:
Thanks for this.

PR #1553, opened 2021-12-24
Remove remaining references to nng_getopt_ptr

### LINKED ISSUE #1071: tran_chkopt can be cleaned up

rettichschnidi:
nng_getopt_ptr got removed/reworked in commit 249932f3a208260f6b9c99d778
[fixes #1071 tran_chkopt can be cleaned up] and since then, had a
declaration but no definition.


gdamore:
Wow, I think those removals were possibly a mistake.  I certainly didn't intend to remove them, although they were largely unused.

gdamore:
Yeah, I'm going to decline this, and instead put back the implementation of nng_getopt_ptr into nng_legacy.c.  The removal of it was a mistake, and in theory could break applications built to use it.  (That said it's probably pretty unlikely that such applications exist -- only TLS as a transport needs it.)

PR #1564, opened 2022-01-01
nni_aio_fini can be simpler.

gdamore:
We don't have to do the close, because all callers should have done
nni_aio_stop first.  This may simplify some of our code paths later
by allowing us to eliminate reap calls that exist solely to permit
clean shut down of aios.


gdamore:
We're not ready for this yet.

I think I need to make a sock_stop function first.

PR #1565, opened 2022-01-12
Add MQTT support for NNG

JaylinYu:
Hereby we presenting a huge pull request to make NNG an MQTT SDK, which I think is a big step for NNG.

As we discussed before, I limited the commit number to 6 plus 3 from upstream/mqtt branch. And we also manage to reorganize the code base to fit with NNG style.

And I know there are some places we are agreed to improve, such as the QoS mechanism in the transport layer, and the way to do connect_cb of MQTT. But I believe it is more efficient to proceed this within a Github pr.


cmake/NNGOptions.cmake
```diff
@@ -76,6 +76,9 @@ mark_as_advanced(NNG_PROTO_RESPONDENT0)
 option (NNG_PROTO_SURVEYOR0 "Enable SURVEYORv0 protocol." ON)
 mark_as_advanced(NNG_PROTO_SURVEYOR0)
 
+option (NNG_PROTO_MQTT_CLIENT "Enable MQTT Client protocol." ON)
```
gdamore, line 79: For now (this release), I'd prefer this to be OFF by default. I want to get some test coverage over this before we start having it used everywhere.  I'd also like this to be noted as EXPERIMENTAL for now.  I'm anticipating making changes here over time, and I want to set expectations accordingly.

docs/man/libnng.3.adoc
```diff
@@ -455,6 +455,36 @@ with TLS support.
 |xref:nng_tls_config_server_name.3tls.adoc[nng_tls_config_server_name()]|set remote server name
 |===
 
+=== MQTT Support
+
+The library support MQTT V3.1.1 for now (V5.0 is being prepared for development)
```
gdamore, line 460: Let's please have an indication that these interfaces are experimental, and may change in future releases as we gain experience with them.

docs/man/nng_mqtt_client_open.3.adoc
```diff
@@ -0,0 +1,70 @@
+= nng_mqtt_client_open(3)
+//
+// Copyright 2018 Staysail Systems, Inc. <info@staysail.tech>
+// Copyright 2018 Capitar IT Group BV <info@capitar.com>
```
gdamore, line 4: You should probably use a different copyright here.

docs/man/nng_mqtt_client_open.3.adoc
```diff
@@ -0,0 +1,70 @@
...
+----
+#include <nng/nng.h>
+#include <nng/mqtt/mqtt_client.h>
+
+int nng_mqtt_client_open(nng_socket *sock);
+----
+
+== DESCRIPTION
+
+The `nng_mqtt_client_open()` Create a MQTT client dose nit connect it.
```
gdamore, line 28: dose is not a word?

src/core/idhash.c
```diff
@@ -61,6 +61,18 @@ nni_id_map_fini(nni_id_map *m)
 	}
 }
 
+void
+nni_id_map_foreach(nni_id_map *m, nni_cb cb)
```
gdamore, line 65: I haven't looked at your usage yet, but I suspect it would be useful to have a different callback signature, one that provided both the id, and the value stored.

src/core/message.c
```diff
@@ -24,11 +24,13 @@ typedef struct {
...
 // Underlying message structure.
 struct nng_msg {
-	uint32_t       m_header_buf[(NNI_MAX_MAX_TTL + 1)];
-	size_t         m_header_len;
-	nni_chunk      m_body;
-	uint32_t       m_pipe; // set on receive
-	nni_atomic_int m_refcnt;
+	uint32_t           m_header_buf[(NNI_MAX_MAX_TTL + 1)];
+	size_t             m_header_len;
+	nni_chunk          m_body;
```
gdamore, line 29: I'm not sure I have a better solution at present, but a few things about this design:

1. It increases the size of every message by 2 pointers (often this will be 16 bytes).
2. I think having protocol details in this particular manner may make moving messages between protocols very difficult.  I'm particular concerned about user applications that might want to proxy between e.g. SP PUB/SUB and MQTT.  I'd like to think about ways to make this better.  I need to research how you're using this before I can make a concrete suggestion.

src/core/message.c
```diff
@@ -427,6 +432,17 @@ nni_msg_dup(nni_msg **dup, const nni_msg *src)
 	nni_atomic_init(&m->m_refcnt);
 	nni_atomic_set(&m->m_refcnt, 1);
 
+	// clone protocol data if a method was supplied.
+	if (src->m_proto_ops != NULL && src->m_proto_ops->msg_free != NULL) {
```
gdamore, line 436: My instinct is that this should be per option, rather than per protocol.  See my comments above.

src/core/stream.c
```diff
@@ -44,6 +44,36 @@ static struct {
...
+	    .dialer_alloc   = nni_tcp_dialer_alloc,
+	    .listener_alloc = nni_tcp_listener_alloc,
+	},
+	{
+	    .scheme         = "tls+mqtt-tcp",
+	    .dialer_alloc   = nni_tls_dialer_alloc,
+	    .listener_alloc = nni_tls_listener_alloc,
+	},
+	{
+	    .scheme         = "tls+mqtt-tcp4",
```
gdamore, line 68: I suspect you should place mqtt+ in the front of your URL.  But really I don't think you need this at all.  Your mqtt URLs are not adding any unique functionality over just tls or tcp.

src/mqtt/protocol/mqtt/mqtt_client.c
```diff
@@ -0,0 +1,1110 @@
...
+mqtt_send_cb(void *arg)
+{
+	mqtt_pipe_t *p = arg;
+	mqtt_sock_t *s = p->mqtt_sock;
+	work_t *     work;
+	bool         scheduler = false;
+
+	nni_mtx_lock(&s->mtx);
+
+	work = nni_list_first(&s->send_queue); // will not be NULL
```
gdamore, line 498: Are you really sure that's the case?  It got flagged for a reason.  Most callbacks need to deal with e.g. cancellation.. what happens if this operation is canceled?  Will that clear the send_queue item?

src/mqtt/protocol/mqtt/mqtt_client.c
```diff
@@ -0,0 +1,1110 @@
...
+	if (nni_aio_result(&p->send_aio) != 0) {
+		// We failed to send... clean up and deal with it.
+		nni_msg_free(nni_aio_get_msg(&p->send_aio));
+		nni_aio_set_msg(&p->send_aio, NULL);
+		nni_mtx_unlock(&s->mtx);
+		nni_pipe_close(p->pipe);
+		return;
+	}
+
+	if (nni_atomic_get_bool(&s->closed) ||
```
gdamore, line 510: Suspect these changes should not be needed -- probably the callback would be canceled with NNG_ECLOSED in that case.

src/mqtt/protocol/mqtt/mqtt_client.c
```diff
@@ -0,0 +1,1110 @@
...
+
+typedef nni_mqtt_packet_type packet_type_t;
+
+// Work state indicating what MQTT packet we *expect* to send/recv.
+//
+// bits:   7 6 5 4 3 2 1 0
+// error <_/ | | | |
+// final <___/ | | \________> lower nib encoding packet type
+// init  <_____/ \__________> ack received
+typedef uint8_t work_state_t;
```
gdamore, line 60: I'm not a fan of binary encoding this here.  It makes it harder to debug.

src/mqtt/protocol/mqtt/mqtt_client.c
```diff
@@ -0,0 +1,1110 @@
...
+	work_t *     work;
+
+	nni_mtx_lock(&s->mtx);
+
+	nni_msg *msg = nni_aio_get_msg(&p->recv_aio);
+
+	if (nni_aio_result(&p->recv_aio) != 0) {
+		nni_msg_free(msg);
+		nni_mtx_unlock(&s->mtx);
+		nni_pipe_close(p->pipe);
```
gdamore, line 665: nni_pipe_ etc. are probably going to be reassociated with SP.  The details of transports are fairly SP specific, and I really want to move this work off of using the SP methods.  I know that we have some work to do to finish enabling that.  Just be advised that this will need to change in the future.

src/mqtt/protocol/mqtt/mqtt_client.c
```diff
@@ -0,0 +1,1110 @@
...
+	switch (packet_type) {
+	case NNG_MQTT_CONNACK:
+		// we have received the CONNACK
+		nni_mtx_unlock(&s->mtx);
+		return;
+
+	case NNG_MQTT_PUBACK:
+		// we have received a PUBACK, successful delivery of a QoS 1
+
+		// fall through
```
gdamore, line 697: I think we should say "FALLTHROUGH" -- I expect some checkers are looking for that specific text.

src/mqtt/protocol/mqtt/mqtt_client.c
```diff
@@ -0,0 +1,1110 @@
...
+	work_t *     work = nni_list_first(&s->recv_queue);
+	mqtt_pipe_t *p    = s->mqtt_pipe;
+	nni_msg *    msg;
+
+	while (NULL != work) {
+		if (0 != nni_lmq_get(&p->recv_messages, &msg)) {
+			break;
+		}
+		nni_list_remove(&s->recv_queue, work);
+		// nni_pipe_recv(p->pipe, &work->recv_aio);
```
gdamore, line 891: This looks fishy.   Do you need to restart recv, or not?

src/mqtt/protocol/mqtt/mqtt_client.c
```diff
@@ -0,0 +1,1110 @@
...
+				// fall through
+			case NNG_MQTT_SUBSCRIBE:
+			case NNG_MQTT_UNSUBSCRIBE:
+				work->packet_id =
+				    mqtt_pipe_get_next_packet_id(p);
+				nni_mqtt_msg_set_packet_id(
+				    work->msg, work->packet_id);
+				NNI_ASSERT(nni_id_get(&p->send_unack,
+				               work->packet_id) == NULL);
+				if (0 !=
```
gdamore, line 944: Style:  I *hate* this style.  Please put the constant *after* the condition.  And if this is a boolean condition, please just use boolean tests, rather than comparing against integers.

gdamore (changes requested):
I'm not done with the review... but this is a start.

JaylinYu:
> 

roger that. Glad you start looking at our PR. the start is the beginning of the end.

src/core/message.c
JaylinYu: We did not touch any of the original parts of nng_msg, just follow your design in MQTT branch.

src/core/message.c
JaylinYu: This is from your code.... I was having an effort to persuade you to do so, perhaps we need another talk.

src/core/stream.c
```diff
@@ -44,6 +44,36 @@ static struct {
 	    .listener_alloc = nni_ipc_listener_alloc,
 	},
 #endif
```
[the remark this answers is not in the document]
JaylinYu: To make NNG MQTT compatible is our goal, I don't think it will mislead users to use SP with MQTT, as long as we provide clear and straight docs. However, we can make a totally independent URL translator and dialer/initiator for MQTT, but needs your hands and help to do it.

src/mqtt/protocol/mqtt/mqtt_client.c
JaylinYu: roger that. I am using MQTT specific transport layer for now.

cmake/NNGOptions.cmake
JaylinYu: done

src/core/idhash.c
JaylinYu: done

src/mqtt/protocol/mqtt/mqtt_client.c
JaylinYu: just rewrite whole file. please have a look again. Any comment is appreciated.

src/core/message.c
JaylinYu: And I am really interested on proxy between pub/sub of SP and MQTT, could you give me more detail on this? Maybe I can implement this feature at first.

docs/man/libnng.3.adoc
JaylinYu: will do

docs/man/nng_mqtt_client_open.3.adoc
JaylinYu: gonna fix it

src/core/stream.c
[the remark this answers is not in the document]
JaylinYu: I do not quite understand what you want exactly ..., perhaps just rename it? I prefer to let users share same entrance just as SP when using nng for mqtt.

src/core/stream.c
JaylinYu: Ok, as your wish. whatever you name it ^_^

src/mqtt/protocol/mqtt/mqtt_client.c
JaylinYu: I am not sure either. lack a way to identify it.

src/mqtt/protocol/mqtt/mqtt_client.c
JaylinYu: done

src/mqtt/protocol/mqtt/mqtt_client.c
JaylinYu: done

src/mqtt/protocol/mqtt/mqtt_client.c
JaylinYu: removed

src/core/message.c
gdamore: It would be pretty easy to take messages from an SP SUB and forward them to MQTT as publish messages.  It would be also easy to take incoming MQTT messages and turn them into SP PUB.  Of course QoS and some details like topic are going to get lost or need to be handled specially.

We should try to avoid burdening messages in a way that impedes this kind of usage.  If the message needs to convey protocol specific data (outside of the headers which wind up being "hidden" from applications when using "cooked" sockets) -- particularly protocol-specific data that needs to be conveyed all the way to the application -- then we should do this in a way so that the content can be trivially ignored by other protocols, and that other protocols can "add" their own protocol-specific data.

One good way to do this, IMO, is to locate options on the message using a key value pair.  It would also be possible to use dynamic IDs (sort of like some thread-specific-data schemes do), where we register a symbolic key, and get back a numeric ID, which might just become an array index.  (Although I suspect a simple key-value pair with a linked list is probably sufficient for most uses -- it might have a very minor impact on performance, but I'd think that walking a very short linked list of < 5 nodes would not do too much to seriously impact the performance.)

The details of how the data are attached to the message should be hidden behind an API however, so that we can adjust the details according to performance needs without breaking application consumers.

src/core/message.c
gdamore: Possibly.  I retain the prerogative to change my mind as I think more about these problems. :-)

src/core/stream.c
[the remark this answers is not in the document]
gdamore: The issue here is that these URLs describe *transports* (stream protocols).  You should not declare new ones here.

MQTT should for example just use tcp:// or tls+tcp:// or whatever in its URLs.  No need for an MQTT prefix.

The bigger problem here is that you really shouldn't be using the low level SP listener and dialer framework.  (In fact, for MQTT, being a listener is non-sensical since only an MQTT broker should listen.)

I've been working on moving the listener/dialer/pipe/socket semantics that are SP specific into the SP directory, and exposing only some very very high level common semantics (using an object oriented design) at the common level.

src/core/stream.c
gdamore: Yeah, in retrospect, mqtt specific URLs are not needed.  See above.

src/core/stream.c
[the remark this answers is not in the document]
JaylinYu: Well, MQTT requires an independent transports protocol as well. I believe we discussed this before, about how to implement QoS while keep zero-copy of nng.
Anyway, what would you want me to do with URL? How about I duplicate the whole set of listener and dialer framework for mqtt first?

brugr9:
Dear @JaylinYu or @gdamore 
May I ask nicely: Is there any news if it is possibly in planning that this request will be completed?
Thanks for your service.
Kindest regards,
Roland Bruggmann aka @brugr9

JaylinYu:
new one is coming

PR #1595, opened 2022-05-31
Fix out-of-bounds error in http_uri_canonify().

### LINKED ISSUE #1594: http_uri_canonify() out-of-bounds
http_uri_canofify() fails for URIs terminating with e.g. %AF.
It goes out of bounds.

** Environment Details **
 - NNG version 1.5.2 (and previous version)
 - Operating system and version Windows / Linux
 - Compiler and language used. clang / msvc 
 - Both shared and static libraries

manuelgustavo:
fixes #1594 undefined behaviour due to out-of-bounds in http_uri_canonify().

A small, but important change was done in http_uri_canonify() to mend the issue.

### DIFF
--- a/src/supplemental/http/http_server.c
+++ b/src/supplemental/http/http_server.c
@@ -455,10 +455,11 @@ http_uri_canonify(char *path)
 			c += http_hexval(tmp[2]);
 			*dst++ = c;
 			tmp += 3;
+		} else {
+			// garbage in, garbage out
+			*dst++ = c;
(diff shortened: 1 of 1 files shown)


gdamore:
It looks good, thanks for catching this.  Probably it should be a test case somewhere.

manuelgustavo:
Cheers... It's undefined behaviour... So it won't always crash.
And thanks for developing the library!

Best,
Manuel

PR #1599, opened 2022-06-18
fixes non-finishing aio when CancelIoEx fails (#1598)

### LINKED ISSUE #1598: http_server deadlocks on shutdown
**Describe the bug**
Probably is not a bug, but an issue on my code, but I cannot figure out what should be done.

So, before calling it a bug, I would like to know if the shutdown code is correct, or I missed some other function.

**Expected behavior**
After receiving a particular GET query, the server must shut down.

**Actual Behavior**
The server deadlocks in `nng_http_server_stop`

** Environment Details **
 - latest master (41ce9fbfd7a396816dab08dcb0284bab4329c622)
 - Windows 10
 - C++
 - Static library

**Additional context**
The issue is a lot more visible when using the VPN client `AstrillVPN`. 
With it, it happens like 80% of the times, even if it is not enabled (just being installed).

**To Reproduce**

The following code deadlocks. 
All error handling has been removed for clarity. No errors are thrown.

The code deadlocks when calling `nng_http_server_stop()`.
It hangs waiting for closing of all connections (in `http_server_stop()`, line `nni_cv_wait(&s->cv);` )

While the reaper thread trying to close the connections is waiting for completion of the task on `conn->rd_aio`
(in `nni_http_conn_fini()`, line `nni_aio_stop(conn->rd_aio);`

Thank you for any input you can provide, as I am out of ideas about what to do here....
Best regards,
Ruben

```
#include <nng/nng.h>
#include <nng/supplemental/http/http.h>

#include <iostream>
#include <string>
#include <thread>
#include <chrono>
#include <atomic>

std::atomic<bool> callback_done(false);

static void handler_callback(nng_aio *io) {
    // Actual handler work and response generation
    callback_done = true;
    
    nng_http_res *nng_res=NULL;
    nng_http_res_alloc(&nng_res);
    
    // It doesn't matter if it is just a redirection or an actual HTML page.
    
    //nng_http_res_set_status(nng_res, 302);
    //nng_http_res_set_header(nng_res, "Location", "https://www.google.com");
    //nng_http_res_set_header(nng_res, "Connection", "close");
    
    nng_http_res_set_status(nng_res, 200);
    std::string body = "<html><body>Hello World</body></html>";
    nng_http_res_copy_data(nng_res, body.c_str(), body.size());
    
        
    // Return the response
    nng_aio_set_output(io, 0, nng_res);
    
    // And finish aio
    nng_aio_finish(io, 0);
}

int main (int argc, char *argv[]) {
    // Setup server
    // ------------
    
    // Create server
    nng_http_server *server = nullptr;
    {
        nng_url *server_url=nullptr;
        nng_url_parse(&server_url, "http://127.0.0.1:8080");
    
        nng_http_server_hold( &server, server_url);
        nng_url_free(server_url);
    }
    
    // Create and add handler
    nng_http_handler *nng_handler=nullptr;
    nng_http_handler_alloc(&nng_handler, "/callback", &handler_callback);
    nng_http_server_add_handler(server, nng_handler);
        
    // Start server   
    nng_http_server_start(server);
    
    // Normal Server Work
    // ------------------
    
    // Wait for the callback to be called
    while (!callback_done) {
        std::this_thread::sleep_for(std::chrono::milliseconds(20));
    }
    
    // Shutdown server
    // ---------------
    
    // Stop server
    nng_http_server_stop(server); // HERE it deadlocks waiting for completion of the task on conn->rd_aio
    
    // Remove handler
    nng_http_server_del_handler(server, nng_handler);
    
    // Destroy server and handler
    nng_http_server_release(server);
    nng_http_handler_free(nng_handler);
    
    return 0;
}

```

L0PiTaL:
fixes #1598 If CancelIoEx fails, it does not call the read/write callbacks, so the AIOs are never finished, leading to a deadlock when closing a TCP server.

As commented in #1598, while tihs PR is focusing in `win_tcpconn.c`, this same issue may happen in other files.

If you consider this solution is fine, I can "complete" the PR with all the other places where `CancelIoEx` appears, but I wanted to check it with you before making further changes.

### DIFF
--- a/src/platform/windows/win_tcpconn.c
+++ b/src/platform/windows/win_tcpconn.c
@@ -17,6 +17,17 @@
 #include <stdio.h>
 
 static void
+tcp_recv_clear_aios(nni_tcp_conn *c)
+{
+	nni_aio *aio;
+	while ((aio = nni_list_first(&c->recv_aios)) != NULL) {
+		nni_list_remove(&c->recv_aios, aio);
+		nni_aio_finish_error(aio, NNG_ECLOSED);
+	}
+	nni_cv_wake(&c->cv);
+}
+
+static void
 tcp_recv_start(nni_tcp_conn *c)
 {
 	nni_aio *aio;
@@ -29,11 +40,7 @@ tcp_recv_start(nni_tcp_conn *c)
 	WSABUF * iov;
 
 	if (c->closed) {
-		while ((aio = nni_list_first(&c->recv_aios)) != NULL) {
-			nni_list_remove(&c->recv_aios, aio);
-			nni_aio_finish_error(aio, NNG_ECLOSED);
-		}
-		nni_cv_wake(&c->cv);
+		tcp_recv_clear_aios(c);
 	}
 again:
 	if ((aio = nni_list_first(&c->recv_aios)) == NULL) {
@@ -101,7 +108,12 @@ tcp_recv_cancel(nni_aio *aio, void *arg, int rv)
 	nni_mtx_lock(&c->mtx);
 	if (aio == nni_list_first(&c->recv_aios)) {
 		c->recv_rv = rv;
-		CancelIoEx((HANDLE) c->s, &c->recv_io.olpd);
+		if (CancelIoEx((HANDLE) c->s, &c->recv_io.olpd)==0) {
+			// Some error happened. We will not have a callback. Finish AIO
+			nni_aio_list_remove(aio);
+			nni_aio_finish_error(aio, rv);
+			nni_cv_wake(&c->cv);
+		}
 	} else if (nni_aio_list_active(aio)) {
 		nni_aio_list_remove(aio);
 		nni_aio_finish_error(aio, rv);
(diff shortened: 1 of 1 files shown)


src/platform/windows/win_tcpconn.c
```diff
@@ -191,7 +209,12 @@ tcp_send_cancel(nni_aio *aio, void *arg, int rv)
 	nni_mtx_lock(&c->mtx);
 	if (aio == nni_list_first(&c->send_aios)) {
 		c->send_rv = rv;
-		CancelIoEx((HANDLE) c->s, &c->send_io.olpd);
+		if (CancelIoEx((HANDLE) c->s, &c->send_io.olpd) == 0) {
+			// Some error happened. We will not have a callback. Finish AIO
```
gdamore, line 213: I concerned what the semantics here of CancelIoEx are.  If this fails, does it mean that the I/O is still pending?

It would be useful to know the error code... is it ERROR_NOT_FOUND?  If so, why?  Is this happening because cancel is racing against finish?

My biggest worry here is that we are somehow treating a failure to cancel as a statement that the I/O is not active.  If that's absolutely certainly the case, then I'm good with this.  If it isn't (because for example the operation cannot be canceled for some other reason), then this approach is unacceptably risky.

So I'm looking to better understand what the circumstances are when this fails.

gdamore (changes requested):
I'm concerned mostly about the safety of this change.  Collecting GetLastError() when CancelIoEx fails would be really useful.

Perhaps the problem is that we failed to schedule it (cancellation arriving before posting the request to the I/O subsystem?)

Essentially what I want here is a little more analysis to be sure that we're fixing the problem in the right place, and not leaving us with a potential memory corrupting race elsewhere.

L0PiTaL:
@gdamore 
You are right. This PR is not good enough, as I made it crash with a combination of cancel and late data reception.
Closing it.

PR #1600, opened 2022-06-18
Fixes compiling when NNG_ENABLE_STATS is OFF

L0PiTaL:
The socket member `st_name` is not available when NNG_ENABLE_STATS is OFF.


gdamore:
Thanks for fixing this.  I guess I should probably add testing this to the CI/CD.

PR #1601, opened 2022-06-18
Fix mingw atomics undefined

L0PiTaL:
There are two issues with Mingw atomics defines which are solved by this PR's commits:

1. `InterlockedDecrementAcquire64` was not defined, so the build failed on Mingw.
2. `InterlockedDecrementRelease64` was incorrect. The docs for this function state that 
"The function returns the resulting decremented value."
The equivalent function for this in Mingw is `__atomic_sub_fetch`, and not `__atomic_fetch_sub` (which returns the value previous to the decrement).


gdamore:
Oops, thanks for fixing this.  It's unfortunate that mingw needs these special handling cases.

PR #1604, opened 2022-07-13
Fixes the call to setsockopt(SO_UPDATE_ACCEPT_CONTEXT) on Windows (#1598)

### LINKED ISSUE #1598: http_server deadlocks on shutdown

L0PiTaL:
From the [docs](https://docs.microsoft.com/en-us/windows/win32/api/mswsock/nf-mswsock-acceptex), this function requires as a parameter the listener socket, not a boolean value.
With this change, `shutdown` now can start the socket closing process correctly.

fixes #1598

### DIFF
--- a/src/platform/windows/win_tcplisten.c
+++ b/src/platform/windows/win_tcplisten.c
@@ -88,7 +88,6 @@ tcp_accept_cb(nni_win_io *io, int rv, size_t cnt)
 	int               len2;
 	SOCKADDR *        sa1;
 	SOCKADDR *        sa2;
-	DWORD             yes;
 	BOOL              nd;
 	BOOL              ka;
 
@@ -123,9 +122,8 @@ tcp_accept_cb(nni_win_io *io, int rv, size_t cnt)
 	memcpy(&c->sockname, sa1, len1);
 	memcpy(&c->peername, sa2, len2);
 
-	yes = 1;
 	(void) setsockopt(c->s, SOL_SOCKET, SO_UPDATE_ACCEPT_CONTEXT,
-	    (char *) &yes, sizeof(yes));
+	    (char *) &l->s, sizeof(l->s));
 
 	(void) setsockopt(
 	    c->s, SOL_SOCKET, SO_KEEPALIVE, (char *) &ka, sizeof(ka));


src/platform/windows/win_tcpconn.c
```diff
@@ -260,7 +260,13 @@ tcp_close(void *arg)
 	if (!c->closed) {
 		c->closed = true;
 		if (!nni_list_empty(&c->recv_aios)) {
-			CancelIoEx((HANDLE) c->s, &c->recv_io.olpd);
```
gdamore, line 263: So the problem is that closing the socket didn't up the waiting receivers?  Is that the situation?  (A similar problem exists with UNIX and there some unfortunate workarounds for it in the POSIX side of the code.)

L0PiTaL: @gdamore 

FINALLY!
After your comment about closing the socket, I have checked the result of `shutdown()` and it also was raising an error (`WSAENOTCONN`).

After searching this error, it came from an issue on the call to `setsockopt(SOL_SOCKET, SO_UPDATE_ACCEPT_CONTEXT)` after AcceptEx.

From the [docs](https://docs.microsoft.com/en-us/windows/win32/api/mswsock/nf-mswsock-acceptex), this function requires as a parameter the listener socket, not a boolean value.

Now it closes the connections correctly.

I have rebsaed this PR with these new changes, so it should be ready for review.

gdamore: Excellent.  Strange behavior in Windows around this, but clearly what I had was a bug.

PR #1617, opened 2022-09-14
Get common name and subject alternative names of peer certificate

fishbach:
When doing TLS client authentication, the TLS server may need to query the common name of the client.

Example:

```
nng_socket sock;
nng_msg *msg;
nng_pipe p;
char * cn;

nng_recvmsg(sock, &msg, 0);
p = nng_msg_get_pipe(msg);
nng_pipe_get_string(p, NNG_OPT_TLS_PEER_CN, &cn);
printf("peer CN: %s\n", cn);
free(cn);
```


gdamore:
This looks like a very nice contribution.

I need to check because it does extend the engine interface and there is a (not open source) repository that implements the engine API ... so it will probably need to be updated.

Issue #1175, opened 2020-01-23
nng_listen occasionally fails when reusing ipc socket on Windows


jeikabu:
## NNG & Platform details.

nng v1.2.4 Win10

## Expected Behavior

After calling `nng_close()` on a socket listening on `"ipc://"`, can immediately open another socket and listen on same pipe.

## Actual Behavior

Works on Linux.  On Windows, `nng_listen()` fails with `EADDRINUSE` after 10-500x iterations.

## Steps to Reproduce

```c
int
main(int argc, char** argv) {
    nng_socket socket0, socket1;
    int i;
    const char* url = "ipc://test_socket";
    for (int i = 0; i < 1000; i++) {
        assert(nng_pair0_open(&socket0) == 0);
        assert(nng_listen(socket0, url, NULL, 0) == 0);
        assert(nng_pair0_open(&socket1) == 0);
        assert(nng_dial(socket1, url, NULL, 0) == 0);
        assert(nng_close(socket1) == 0);
        assert(nng_close(socket0) == 0);
    }
    
    return 0;
}
```

gdamore:
I'm assuming you're using actual Windows and not WSL.

IPC on Windows is implemented using named pipes.  I wonder if the OS has some limitation here -- I don't think I'm doing anything particularly asynchronously -- but I will look into it.

gdamore:
I suspect that what is happening is that while I'm calling DisconnectNamedPipe -- you might have clients on the far side that are still holding that point in the "namespace" active.  This is different from how it works on UNIX or with TCP.  If this is actually what is happening, I'm not sure I'll be able to do much about it.

gdamore:
Docs about this are here: https://docs.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-disconnectnamedpipe

Essentially you have to wait for the clients to see the pipe error, and close the pipe. 

This does mean that a malicious client can hold the pipe indefinitely, and prevent future server instances from listening.  This is a design defect in Named Pipes in Windows, IMO.  (But again, Windows wasn't designed for multi-user use, really, and this API reflects that -- mostly processes are assumed to be cooperative and not actively hostile.)

jeikabu:
Tried looking into this and this seems to be what's going on:

[...]

PR #1628, opened 2022-11-01
change windows IPC prefix

alzix:
according to MS documentation:
> Windows 10, version 1709:  Pipes are only supported within an app-container; ie, from one UWP process to another UWP process that's part of the same app. Also, named pipes must use the syntax **"\.\pipe\LOCAL"** for the pipe name.

references:
https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createnamedpipea#remarks https://learn.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-disconnectnamedpipe#remarks

### DIFF
--- a/src/platform/windows/win_ipc.h
+++ b/src/platform/windows/win_ipc.h
@@ -17,7 +17,7 @@
 #include "core/nng_impl.h"
 #include "win_impl.h"
 
-#define IPC_PIPE_PREFIX "\\\\.\\pipe\\"
+#define IPC_PIPE_PREFIX "\\\\.\\pipe\\LOCAL\\"
 
 extern int nni_win_ipc_init(nng_stream **, HANDLE, const nng_sockaddr *, bool);
 


gdamore:
That's a new requirement for Windows 10 1709.  I'm not sure I want to adopt this universally as it might break some applications.

gdamore:
Is it possible that the name restriction is specific to UWP apps?

alzix:
i do not know.
i was trying to investigate issue #1175 as it impacts my work - so i read native APIs documentation.
i thought it might be relevant.
i cannot conclude it solves the issue, though. But it seemes to me it did reduce the reproduction frequency.

gdamore:
I don't think this has any particular relevance for #1175 unforutnately.

gdamore:
I'm going to decline this.  Windows applications that need to honor this to work in UWP need to define a suitable path including the LOCAL prefix for now.  This may be a little unfortunate, but I'd rather not break compatibility with existing libraries.

PR #1638, opened 2023-01-19
Fix UBSAN error in idhash.c

shikokuchuo:
Fix for UBSAN error, such as excerpt below:

```
idhash.c:132:14: runtime error: null pointer passed as argument 2, which is declared to never be null
/usr/include/string.h:44:28: note: nonnull attribute specified here
```

As `id_reg_map` is initialised as NULL and passing NULL to `memcpy()` is undefined. Should make no difference to compiled code. Purely to appease the automated checks I have to deal with on my side. Thanks!

### DIFF
--- a/src/core/idhash.c
+++ b/src/core/idhash.c
@@ -129,7 +129,8 @@ id_map_register(nni_id_map *m)
 			return (NNG_ENOMEM);
 		}
 		id_reg_len = len;
-		memcpy(mr, id_reg_map, id_reg_num * sizeof(nni_id_map *));
+		if (id_reg_map != NULL)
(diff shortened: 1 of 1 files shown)


gdamore:
Yeah, this is one of those stupid things -- when we start out we have a NULL value and zero length, so passing *any* value there should be meaningless because we aren't going to copy anything at all (the length is zero).

This is not on a hot code path, so this fix is fine.  If it were on a hot code path we might elect instead to initialize the pointer to something (other than NULL), so that we don't need an extra branch instruction (and the possible stall.) This is probably more readable though, so thanks.

PR #1640, opened 2023-01-30
Fix PAIR v0 not finalizing the pollables when closing socket

jimenezrick:
This function needs to be pretty much identical between PAIR v0 and v1, it was missing just the call to release the pollable resources.

I was testing the compat API and i discovered that when using PAIR v0, it was leaking file descriptors because of this. This solves the problem completely mirroring the same logic present in PAIR v1 (https://github.com/nanomsg/nng/blob/master/src/sp/protocol/pair1/pair.c#L79-L80).


gdamore:
Thanks for the fix!

gdamore:
The leaks in the test suites are unrelated to this.

PR #1650, opened 2023-04-16
fix potential deadlock in nni_reap_drain

alzix:
`reap_empty` must be marked as `volatile` as it changed from another thread while being polled in the `nni_reap_drain`
```C
nni_reap_drain(void)
{
	nni_mtx_lock(&reap_mtx);
	while (!reap_empty) {
		nni_cv_wait(&reap_empty_cv);
	}
	nni_mtx_unlock(&reap_mtx);
}
```
compiler must be aware to not apply any optimizations to the `reap_empty` - otherwise on highly optimized code this `while` will spin forever

### DIFF
--- a/src/core/reap.c
+++ b/src/core/reap.c
@@ -19,7 +19,7 @@ static nni_reap_list *reap_list = NULL;
 static nni_thr        reap_thr;
 static bool           reap_exit = false;
 static nni_mtx        reap_mtx = NNI_MTX_INITIALIZER;
-static bool           reap_empty;
+static volatile bool  reap_empty;
 static nni_cv         reap_work_cv = NNI_CV_INITIALIZER(&reap_mtx);
 static nni_cv         reap_empty_cv = NNI_CV_INITIALIZER(&reap_mtx);
 


gdamore:
This is incorrect.  The use of the lock and the condition variable ensures that it cannot be optimized away.  

If you have reasons to believe that this isn't working for some reason, please provide some corroborating evidence. 

As someone who works frequently in device drivers and multithreaded kernels, I'm well familiar with volatile -- it is most useful in circumstances where value is in a register or something similar, where a read has a side effect or where the hardware may trigger the value asynchronously.  It *could* be used like that in a circumstance here without a thread synchronizer like I have used, but that would almost always represent a serious defect -- as synchronization has to consider far more than just synchronizing at the level instructions.  

The correct way to deal with this is to either use threading primitives (as used here) or atomic operations (which provide cache synchronization).  In fact the threading primitives are built upon both atomics operations and in this particular case also (typically) upon the kernel thread scheduler.

For now I'm closing this.

alzix:
code analysis complains about same value being read in the loop. 

please see disassembly code of the `nni_reap_drain` here: https://godbolt.org/z/j4adqsnja
remove the `-O` flag and see the difference - with the optimization it does not loads and value and does branch unconditionally

gdamore:
Your code analysis is wrong.  An optimizer making this assumption would break a great many multithreaded programs - the approach here is a common paradigm for multithreaded programs and this is almost the canonical example for condition variables. 

A static analyzer without all of code (ie the thing that sets the flag) might correctly infer this can be constant folded.  But not if you have the entire file.

gdamore:
Perhaps you have an incomplete or incorrect implementation of the condition variables though.

alzix:
have you seen the Compiler Explorer example? i used `std::condition_variable` to demonstrate this

alzix:
you are right.
with more context (another thread) the compiler does not optimizes the variable.
https://godbolt.org/z/4Pbje5EW5

PR #1651, opened 2023-04-19
posix_resolv_gai.c: add AI_NUMERICSERV where undefined

barracuda156:
Borrowed from: https://github.com/macports/macports-legacy-support/blob/2806f66dbd0a7a47fb97fda5f9caedbac4b5f18b/include/netdb.h#L28-L30

### DIFF
--- a/src/platform/posix/posix_resolv_gai.c
+++ b/src/platform/posix/posix_resolv_gai.c
@@ -33,6 +33,10 @@
 #define NNG_RESOLV_CONCURRENCY 4
 #endif
 
+#ifndef AI_NUMERICSERV
+#define AI_NUMERICSERV 0
+#endif
+
 static nni_mtx  resolv_mtx  = NNI_MTX_INITIALIZER;
 static nni_cv   resolv_cv   = NNI_CV_INITIALIZER(&resolv_mtx);
 static bool     resolv_fini = false;


gdamore:
I'm not sure how much I like this -- AI_NUMERICSERV might not have that value and there might be a collision.  

The macports version is probably fine for a single platform (mac) where the value is known to be safe, but this is used on many other platforms.  I'll look at this a bit more closely to see if we can do something more portable.

barracuda156:
> I'm not sure how much I like this -- AI_NUMERICSERV might not have that value and there might be a collision.
> 
> The macports version is probably fine for a single platform (mac) where the value is known to be safe, but this is used on many other platforms. I'll look at this a bit more closely to see if we can do something more portable.

@gdamore Thank you very much! That would be awesome. (Should I close this PR or wait until the update from your side?)

P. S. I have seen elsewhere (and used that too) `AI_NUMERICSERV` defined to 0 where undefined. Not sure if that is any better though.

gdamore:
Looking at the code, using zero would be far superior, and fine.
It might cause the compiler to complain about a useless "| 0" operation.
The other option would be to put the code that ors that into the hints inside an #ifdef check directly.

barracuda156:
> Looking at the code, using zero would be far superior, and fine. It might cause the compiler to complain about a useless "| 0" operation. The other option would be to put the code that ors that into the hints inside an #ifdef check directly.

Defined to 0, rebased.

gdamore (approved):

PR #1677, opened 2023-08-10
Skip URI parameters when serving a directory (http_server)

robiwano:
When serving a directory, an URI such as `http://127.0.0.1/flutter_service_worker.js?v=666216136` will fail because the HTTP server thinks the URI parameter is part of the filename. This PR drops the URI parameters when serving a directory.


gdamore:
This looks reasonable.  I want to make sure that the sanitize is correct.  Stay tuned.

gdamore (approved):

Issue #1523, opened 2021-11-15
C/C++ executable is SEGVing sporadically in `nni_list_remove`.


WJCFerguson:
## Description

This is not currently very actionable by nng devs as I cannot present a simple way to reproduce it.  But I'm reporting it in case someone can offer any helpful insight as I continue to try to track it down.

I'm using direct C bindings for nng 1.5.2, straightforward modest use of PUB/SUB only so far.  Sadly it's something of a Heisenbug, as instrumenting the code reduces its frequency.

Here are two SEGV stack traces I've observed:

```
#0  0x0000005572ba6240 in nni_list_remove ()
#1  0x0000005572bb6400 in sub0_recv_cb ()
#2  0x0000005572bb0584 in nni_task_exec ()
#3  0x0000005572ba2da0 in nni_aio_finish_impl ()
#4  0x0000005572ba2e20 in nni_aio_finish_sync ()
#5  0x0000005572bb8b10 in ipc_pipe_recv_cb ()
#6  0x0000005572bb0114 in nni_taskq_thread ()
#7  0x0000005572bb0b40 in nni_thr_wrap ()
#8  0x0000005572bb41bc in nni_plat_thr_main ()
#9  0x0000007f88277088 in start_thread (arg=0x7fd0cde9df) at pthread_create.c:463
#10 0x0000007f87e64ffc in thread_start () at ../sysdeps/unix/sysv/linux/aarch64/clone.S:78
```

```
#0  0x0000005570b752d0 in nni_list_remove ()
#1  0x0000005570b836c4 in sub0_ctx_cancel ()
#2  0x0000005570b721d8 in nni_aio_expire_loop ()
#3  0x0000005570b7e994 in nni_thr_wrap ()
#4  0x0000005570b81e1c in nni_plat_thr_main ()
#5  0x0000007f790d8088 in start_thread (arg=0x7fe75ba76f) at pthread_create.c:463
#6  0x0000007f78cc5ffc in thread_start () at ../sysdeps/unix/sysv/linux/aarch64/clone.S:78
```

Since I found no other reports mentioning `nni_list_remove`, I initially suspected it could be issues outside nng, e.g. other code tramping over memory, so I ran under valgrind with our code compiled DEBUG, and found no problems, though the frequency of crashes reduced running our the DEBUG built code.

I have now built nng itself DEBUG, so far no crashes, but execution time has been limited so far.

[...]

PR #1695, opened 2023-10-24
Add a new nni_list_node named a_finish_node in struct nng_aio

### LINKED ISSUE #1682: setting the sub receive socket with nng_socket_set_ms will generates a segment fault
**Describe the bug**
Open a process to send data using “pub”, and two processes to receive data using “sub”. Use “nng_socket_set_ms” to set the timeout for the receiving socket of the “sub” process. After running for a period of time (a few hours), the “sub” process will generate a segmentation fault, while it will not generate a segmentation fault if the timeout is not set.

**Expected behavior**
Both “sub” processes can continuously receive data without generating a segmentation fault.

**Actual Behavior**
Both “sub” processes will generate a segmentation fault.

**To Reproduce**
See this modified demo:
[pubsub.zip](https://github.com/nanomsg/nng/files/12416752/pubsub.zip)
 - ./pilot_pub
 - ./pilot_sub
 - ./pilot_sub

** Environment Details **
 - NNG version 1.5.2 
 - Allwinner T507(Quad A53)
 - aarch64-linux-gnu-gcc(gcc version 7.4.1 20181213)  and C
 - Shared library(libnng.so.1.5.2)

**Additional context**
Open “ulimit -c unlimited” and modify “/proc/sys/kernel/core_pattern” (“/root/nng/core-%e-%p”) to generate core dump files. Use “aarch64-linux-gnu-gdb” for debugging and see that there is an issue with the “nng:task/nng:aio:expire” function calling “nni_list_remove”. “nng:task/nng:aio:expire” is the problem encountered by the two “sub” processes respectively.

willwu1217:
Assume that sub0_ctx_cancel and sub0_recv_cb are executed concurrently.

When sub0_recv_cb is unlocked, and before aio is deleted from the finish queue. nni_list_active in sub0_ctx_cancel may mistakenly believe that aio is still on recv_queue, thus triggering repeated execution of nni_list_remove. The second execution will trigger a null pointer access.

So unlocking  after aio is deleted from the finish list.

fixes #1682  <issue synopsis>

contributing guidelines.

### DIFF
--- a/src/core/aio.h
+++ b/src/core/aio.h
@@ -209,6 +209,7 @@ struct nng_aio {
 	nni_aio_expire_q *a_expire_q;
 	nni_list_node     a_expire_node; // Expiration node
 	nni_reap_node     a_reap_node;
+	nni_list_node     a_finish_node; // Finished node
 };
 
 #endif // CORE_AIO_H
--- a/src/sp/protocol/pubsub0/sub.c
+++ b/src/sp/protocol/pubsub0/sub.c
@@ -326,7 +326,7 @@ sub0_recv_cb(void *arg)
 		return;
 	}
 
-	nni_aio_list_init(&finish);
+	NNI_LIST_INIT(&finish, nng_aio, a_finish_node);
 
 	msg = nni_aio_get_msg(&p->aio_recv);
 	nni_aio_set_msg(&p->aio_recv, NULL);
@@ -388,6 +388,7 @@ sub0_recv_cb(void *arg)
 			nni_pollable_raise(&sock->readable);
 		}
 	}
+	nni_mtx_unlock(&sock->lk);
 
 	// NB: This is slightly less efficient in that we may have
 	// created an extra copy in the face of e.g. two subscriptions,
@@ -404,7 +405,6 @@ sub0_recv_cb(void *arg)
 		nni_list_remove(&finish, aio);
 		nni_aio_finish_sync(aio, 0, len);
 	}
-	nni_mtx_unlock(&sock->lk);
 
 	nni_pipe_recv(p->pipe, &p->aio_recv);
 }
--- a/src/sp/protocol/pubsub0/sub.c
+++ b/src/sp/protocol/pubsub0/sub.c
@@ -388,7 +388,6 @@ sub0_recv_cb(void *arg)
 			nni_pollable_raise(&sock->readable);
 		}
 	}
-	nni_mtx_unlock(&sock->lk);
 
 	// NB: This is slightly less efficient in that we may have
 	// created an extra copy in the face of e.g. two subscriptions,
@@ -405,6 +404,7 @@ sub0_recv_cb(void *arg)
 		nni_list_remove(&finish, aio);
 		nni_aio_finish_sync(aio, 0, len);
 	}
+	nni_mtx_unlock(&sock->lk);
 
 	nni_pipe_recv(p->pipe, &p->aio_recv);
 }


gdamore (approved):
I don't think this will work in all circumstances.  In particular I'm worried about a deadlock situation when the callback on the aio tries to resubmit the aio for receive.  This is actually a fairly common idiom.

I think in that case, you'd wind up with an incorrect attempt to reenter the socket lock, resulting in a deadlock.

The fact that you've located the root cause here (the fact that nni_list_active is misreporting here) is super useful.  I'm going to think about this because clearly that test is busted in this case -- your excellent sleuthing suggests several other solutions might be possible.

willwu1217:
I have updated my PR.
Restored unlock position.
Add a new nni_list_node named a_finish_node in struct nng_aio to keep finished aio. To avoid misreporting of nni_list_active in sub0_ctx_cancel

gdamore:
I need time to assess this properly -- sorry I've been busy.  My two areas of particular concern, which I need to satisfy myself are:

1. Do other protocols have the same defect (it seems likely that they do)
2. If they don't (and maybe even if they do), would it be better to locate this somewhere else.... 

I'm also concerned that increasing the size of the aio will have a deleterious affect for some uses.  There are situations where people have huge numbers of aios, and have complained about how big they are.  This increases the size by up to 16 bytes.  It might be able to make a much smaller increase, or even no increase at all, by utilizing a flag.

Still, I think your work here is good, and I might wind up merging it all, but I want to have time to consider these issues first.

ayizhi:
> I need time to assess this properly -- sorry I've been busy. My two areas of particular concern, which I need to satisfy myself are:
> 
> 1. Do other protocols have the same defect (it seems likely that they do)
> 2. If they don't (and maybe even if they do), would it be better to locate this somewhere else....
> 
> I'm also concerned that increasing the size of the aio will have a deleterious affect for some uses. There are situations where people have huge numbers of aios, and have complained about how big they are. This increases the size by up to 16 bytes. It might be able to make a much smaller increase, or even no increase at all, by utilizing a flag.
> 
> Still, I think your work here is good, and I might wind up merging it all, but I want to have time to consider these issues first.

Anticipating a prompt release

gdamore:
Sorry been a bit busy.  I've been thinking about this problem though -- and one idea I have is to use the existing node as a "singly linked" node, in a way that nni_list_node_active won't think is active.  We can do this because we don't have any requirement for random access (or even random removal) for the completion queue.  So we only need one of the list pointers.

gdamore:
I'm closing this in favor of #1523  -- I think what I've done there is superior, but it was inspired by the excellent work you did here, and I've given what credit I can.

willwu1217:
> I'm closing this in favor of #1523 -- I think what I've done there is superior, but it was inspired by the excellent work you did here, and I've given what credit I can.

I have got your idea. Using an existing node as a "single link" node is indeed a better choice in terms of memory usage and efficiency.  Since reap will only appear after finish， we can ensure that aio completions and aio reap will not execute concurrently. 

One more little suggestion, maybe we can use an anonymous union in struct nng_aio.
```
union {
    nni_reap_node     a_reap_node;
    nni_completions_node     a_completions_node;
};
```

gdamore:
I am not sure that anonymous unions work on all C versions we need to support. I don't think it adds enough value to have them to be worth the pain their use might cause if they aren't.

gdamore:
Also btw I went ahead merged my PR.  I think that there are other places in the code that suffer the same defect and I will try to find and fix them all.

PR #1700, opened 2023-11-09
Add dynamic functionality to lmq

### LINKED ISSUE #1699: Demand for dynamic size event queue
**Is your feature request related to a problem? Please describe.**
When the event queue receives a large amount of events, a lot of those events are lost.

**Describe the solution you'd like**
We would like a configurable queue size, instead of the 16 built in size. If 0 is given as size it's size should be dynamic.

**Describe alternatives you've considered**
lmq has a resize function that can be used to implement the dynamic functionality. Size can be set from cmake.

AdamNusser:
fixes #1699  Demand for dynamic size event queue

Added NNG_SENDBUF_SIZE parameter to set event queue size. When then lmq cap is set to 0 it works with dynamic size. 
lmq_get and lmq_put change allocated size if needed.

### DIFF
--- a/src/core/lmq.c
+++ b/src/core/lmq.c
@@ -14,6 +14,8 @@
 // performance sensitive contexts.  Locking must be done by the caller.
 
 
+#define LMQ_DYNAMIC_ALLOC_SIZE 16
+
 // Note that initialization of a queue is guaranteed to succeed.
 // However, if the requested capacity is larger than 2, and memory
 // cannot be allocated, then the capacity will only be 2.
@@ -29,7 +31,11 @@ nni_lmq_init(nni_lmq *lmq, size_t cap)
 	lmq->lmq_msgs = lmq->lmq_buf;
 	lmq->lmq_cap = 2;
 	lmq->lmq_mask = 0x1; // only index 0 and 1
-	if (cap > 2) {
+	lmq->lmq_dyna = (cap == 0);
+	if (cap == 0) {
+		(void) nni_lmq_resize(lmq, LMQ_DYNAMIC_ALLOC_SIZE);
+	}
+	else if (cap > 2) {
 		(void) nni_lmq_resize(lmq, cap);
 	} else {
 		lmq->lmq_cap = cap;
@@ -81,7 +87,7 @@ nni_lmq_cap(nni_lmq *lmq)
 bool
 nni_lmq_full(nni_lmq *lmq)
 {
-	return (lmq->lmq_len >= lmq->lmq_cap);
+	return (lmq->lmq_dyna ? false : lmq->lmq_len >= lmq->lmq_cap);
 }
 
 bool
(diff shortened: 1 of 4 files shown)


src/core/lmq.c
```diff
@@ -29,7 +31,11 @@ nni_lmq_init(nni_lmq *lmq, size_t cap)
 	lmq->lmq_msgs = lmq->lmq_buf;
 	lmq->lmq_cap = 2;
 	lmq->lmq_mask = 0x1; // only index 0 and 1
-	if (cap > 2) {
+	lmq->lmq_dyna = (cap == 0);
+	if (cap == 0) {
+		(void) nni_lmq_resize(lmq, LMQ_DYNAMIC_ALLOC_SIZE);
+	}
```
gdamore, line 37: style nit, prefer nestling like `} else if (...) {`

gdamore (commented):
I have rather serious concerns about making lmq's able to resize dynamically with lmq_put(). 

The design for this was very specifically meant not to support this, to avoid unbounded memory growth, and to ensure back-pressure that is required for some of the protocols to work.

(We used to have another data structure, mq, that could be dynamic, but we replaced it with lmq for performance reasons.)

Can you provide an explanation of the use case where you think dynamic lmq's help your application.

AdamNusser:
We have an issue when too many events come in within a short time. This causes some events to become discarded. To solve this we would like the dynamic queue, and I believe you have already discussed this with our architect (Lajos Szabados, Rolling Wireless) on discord.

AdamNusser:
As to the unbounded memory growth:
This certainly can be an issue, if large amounts of events come in faster than out, but this feature is purely optional, and if not enabled, works exactly like previously.

gdamore:
I guess I'm still not understanding why you don't just increase the base capacity to begin with.  Making the lmq *dynamic* seems like a recipe for complicated bugs.  If you can't process the messages you have in your queue in a timely fashion, then you have a problem where you're either going to run out of memory, or have a long latency for the messages delivered at the tail of the queue.

So I'd highly recommend just statically tuning the queue size to accommodate your largest anticipated burst.

src/sp/protocol/pubsub0/pub.c
```diff
@@ -73,7 +73,11 @@ pub0_sock_init(void *arg, nni_sock *ns)
 	nni_pollable_init(&sock->sendable);
 	nni_mtx_init(&sock->mtx);
 	NNI_LIST_INIT(&sock->pipes, pub0_pipe, node);
```
gdamore, line 75: This is already a *runtime* tunable via NNG_OPT_SENDBUF....
I don't think I agree with making this a compile time setting -- if you want to change it in your application just call 
`nng_sock_set_int(sock, NNG_OPT_SENDBUF, size);`

gdamore:
I don't believe that this change is desirable.  I really, really don't want to add the complexity and risk of automatically growing lmqs ... it's almost certainly misguided.   I also don't agree with changing this via a compile time setting -- and btw the reasonable defaults are different for different protocols.  So the single global default is IMO undesirable.

Instead, you should set the NNG_OPT_SENDBUF to a size that is large enough to accommodate your largest anticipated burst.

One important thing is that the pub protocol tries to discard *older* messages in favor of newer ones.  The growing LMQ compromises that, and would potentially wind up delivering very stale updates (depending on how far behind the queue got.)

While I appreciate your effort and contribution, I'm going to close this PR as I just don't think this is a good direction to take.

PR #1710, opened 2023-11-26
req: use nni_aio_completions - this is similar to #1523

gdamore:

contributing guidelines.


gdamore:
I strongly suspect that all users of nni_aio_finish_sync() (with the exception of nni_aio_completions_run of course) suffer the same defect.

PR #1711, opened 2023-11-26
fixes #1702 segfault canceling req receive while sending

### LINKED ISSUE #1702: Segfault in req0_ctx_cancel_send()
**Describe the bug**
When developing the R binding, I came across and reported this (on Discord) a few months ago. I have now traced this with the help of valgrind.

**To Reproduce**
1. Start a req socket
2. Start a listener on the socket (no connection)
3. Allocate 2 aios, sending a message on one using `nng_send_aio()` and receiving on the other using `nng_recv_aio()` with an expiry timeout.
4. Wait until the recv aio has expired.
5. Call `nng_aio_free()` on the send aio.

The following segfault happens.

==213611== Invalid write of size 8
==213611==    at 0x4A8B0E84: nni_msg_header_clear (message.c:623)
==213611==    by 0x4A8C4C3C: req0_ctx_cancel_send (req.c:627)
==213611==    by 0x4A8AC7B8: nni_aio_fini (aio.c:138)
==213611==    by 0x4A8AC7EA: nni_aio_free (aio.c:161)
==213611==    by 0x4A8AC7EA: nni_aio_free (aio.c:158)
==213611==    by 0x4A89A194: saio_finalizer (aio.c:304)
==213611==    by 0x4A89A194: saio_finalizer (aio.c:299)
==213611==    by 0x49EF5DC: R_RunWeakRefFinalizer (in /usr/lib/R/lib/libR.so)
==213611==    by 0x49EF83B: ??? (in /usr/lib/R/lib/libR.so)
==213611==    by 0x49F2A54: ??? (in /usr/lib/R/lib/libR.so)
==213611==    by 0x4993264: ??? (in /usr/lib/R/lib/libR.so)
==213611==    by 0x49ADCFF: Rf_eval (in /usr/lib/R/lib/libR.so)
==213611==    by 0x49AFD95: ??? (in /usr/lib/R/lib/libR.so)
==213611==    by 0x49B0BC4: Rf_applyClosure (in /usr/lib/R/lib/libR.so)
==213611==  Address 0x40 is not stack'd, malloc'd or (recently) free'd
==213611== 

 *** caught segfault ***
address 0x40, cause 'memory not mapped'

** Environment Details **
 - NNG 1.6.0pre ([a54820f](https://github.com/nanomsg/nng/commit/a54820ff0e1b74554c7f649e8386ee8c4ecd98f5))
 - Ubuntu 22.04
 - gcc 11.4.0
 - Static library

**Additional context**

As it is clear that `nni_msg_header_clear()` was the culprit, I have removed these 2 lines in `req0_ctx_cancel_send()`:
https://github.com/nanomsg/nng/blob/a54820ff0e1b74554c7f649e8386ee8c4ecd98f5/src/sp/protocol/reqrep0/req.c#L626-L627

which eliminates this segfault.

I do not see any impact on my unit tests, valgrind does not show any memory leaks. Also in real world examples, where the connection is lost, the req resend is still happening.

So I am wondering if it is generally safe to take those out, or in what context those would be needed (the comment says restore the message back to the aio?).

From a binding perspective, the aios are individually allocated for messages and hence can be expected to be short-lived.

gdamore:

contributing guidelines.

### DIFF
--- a/src/sp/protocol/reqrep0/req.c
+++ b/src/sp/protocol/reqrep0/req.c
@@ -537,6 +537,23 @@ req0_ctx_cancel_recv(nni_aio *aio, void *arg, int rv)
 	req0_sock *s   = ctx->sock;
 
 	nni_mtx_lock(&s->mtx);
+
+	// So it turns out that some users start receiving before waiting
+	// for the send notification.  In this case if receiving is
+	// canceled before sending completes, we need to restore the
+	// message for the user.  It's probably a mis-design if the user
+	// is trying to receive without waiting for sending to complete, but
+	// it was reported in the field.  Users who want to avoid this mess
+	// should just start receiving from the send completion callback.
+	if (ctx->send_aio != NULL) {
+		nni_aio_set_msg(ctx->send_aio, ctx->req_msg);
+		nni_msg_header_clear(ctx->req_msg);
+		ctx->req_msg = NULL;
+		nni_aio_finish_error(ctx->send_aio, NNG_ECANCELED);
+		ctx->send_aio = NULL;
+		nni_list_remove(&s->send_queue, ctx);
+	}
+
 	if (ctx->recv_aio == aio) {
 		ctx->recv_aio = NULL;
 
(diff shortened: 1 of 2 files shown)


gdamore:
Arguably we should instead of just returned NNG_ESTATE if the send aio was not complete yet, but I can see how some folks might like to have both operations pending at once.

PR #1725, opened 2023-12-16
fixes #1673 High CPU usage with TLS

### LINKED ISSUE #1673: High CPU usage with TLS
We're building and running a server application on both Linux and Windows which accepts a single PAIR1 client with TLS enabled via mbedtls.

Minimal data is being sent while profiling - just periodic keep-alive messages.

We are running this commit from earlier this year, at the recommendation of @shikokuchuo from another thread (https://github.com/nanomsg/nng/issues/1661):

  commit 8e1836f57e8bcdb228dd5baadc71dfbf30b544e0
  Date:   Thu Feb 9 00:48:17 2023 +0100

On Linux, we see unexpectedly high CPU usage (20-30% single core) in NNG, given the small workload.  Windows CPU usage is near 0 for the same server code and use case.  The gprof output for Linux is below.

Is there a known performance problem as of this older commit?  Or do you have any suggestions for how to investigate further?  Thank you for your time.

NNG version: 8e1836f57e8bcdb228dd5baadc71dfbf30b544e0
Operating system and version: 5.15.0-1028-aws
Compiler and language used: clang / C++
Shared or static library: static

gdamore:
The aio for connections was meant to have an infinite sleep (no timeout), but was getting an initial value of zero, so we were spinning on accept.

### DIFF
--- a/src/supplemental/tls/tls_common.c
+++ b/src/supplemental/tls/tls_common.c
@@ -867,6 +867,7 @@ tls_alloc(tls_conn **conn_p, nng_tls_config *cfg, nng_aio *user_aio)
 	nni_aio_list_init(&conn->send_queue);
 	nni_aio_list_init(&conn->recv_queue);
 	nni_mtx_init(&conn->lock);
+	nni_aio_set_timeout(&conn->conn_aio, NNG_DURATION_INFINITE);
 	nni_aio_set_timeout(&conn->tcp_send, NNG_DURATION_INFINITE);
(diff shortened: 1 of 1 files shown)


gdamore:
If this passes all the tests, which I expect at this point it will, I'll merge this even though it will close the bug.  If the bug isn't actually resolved (I'm pretty sure looking at this it will fix it), then we should reopen it (or open a new one).

PR #1727, opened 2023-12-17
copy and modify tcp file structure

patmaddox:

contributing guidelines.

### DIFF
--- a/include/nng/nng.h
+++ b/include/nng/nng.h
@@ -119,6 +119,11 @@ typedef struct nng_aio  nng_aio;
 // sockaddr, but we have our own to cope with our unique families, etc.
 // The details of this structure are directly exposed to applications.
 // These structures can be obtained via property lookups, etc.
+struct nng_sockaddr_file_descriptor {
+	uint16_t sa_family;
+	int      sa_fd;
+};
+
 struct nng_sockaddr_inproc {
 	uint16_t sa_family;
 	char     sa_name[NNG_MAXADDRLEN];
@@ -162,24 +167,26 @@ struct nng_sockaddr_storage {
 	uint64_t sa_pad[16];
 };
 
-typedef struct nng_sockaddr_inproc   nng_sockaddr_inproc;
-typedef struct nng_sockaddr_path     nng_sockaddr_path;
-typedef struct nng_sockaddr_path     nng_sockaddr_ipc;
-typedef struct nng_sockaddr_in       nng_sockaddr_in;
-typedef struct nng_sockaddr_in6      nng_sockaddr_in6;
-typedef struct nng_sockaddr_zt       nng_sockaddr_zt;
-typedef struct nng_sockaddr_abstract nng_sockaddr_abstract;
-typedef struct nng_sockaddr_storage  nng_sockaddr_storage;
+typedef struct nng_sockaddr_file_descriptor nng_sockaddr_file_descriptor;
+typedef struct nng_sockaddr_inproc          nng_sockaddr_inproc;
+typedef struct nng_sockaddr_path            nng_sockaddr_path;
+typedef struct nng_sockaddr_path            nng_sockaddr_ipc;
+typedef struct nng_sockaddr_in              nng_sockaddr_in;
+typedef struct nng_sockaddr_in6             nng_sockaddr_in6;
+typedef struct nng_sockaddr_zt              nng_sockaddr_zt;
+typedef struct nng_sockaddr_abstract        nng_sockaddr_abstract;
+typedef struct nng_sockaddr_storage         nng_sockaddr_storage;
 
 typedef union nng_sockaddr {
-	uint16_t              s_family;
-	nng_sockaddr_ipc      s_ipc;
-	nng_sockaddr_inproc   s_inproc;
-	nng_sockaddr_in6      s_in6;
-	nng_sockaddr_in       s_in;
-	nng_sockaddr_zt       s_zt;
-	nng_sockaddr_abstract s_abstract;
-	nng_sockaddr_storage  s_storage;
+	uint16_t                     s_family;
+	nng_sockaddr_file_descriptor s_file_descriptor;
+	nng_sockaddr_ipc             s_ipc;
+	nng_sockaddr_inproc          s_inproc;
+	nng_sockaddr_in6             s_in6;
+	nng_sockaddr_in              s_in;
+	nng_sockaddr_zt              s_zt;
+	nng_sockaddr_abstract        s_abstract;
+	nng_sockaddr_storage         s_storage;
 } nng_sockaddr;
 
 enum nng_sockaddr_family {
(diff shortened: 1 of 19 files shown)


include/nng/transport/file_descriptor/file_descriptor.h
```diff
@@ -0,0 +1,30 @@
...
+#ifndef NNG_TRANSPORT_TCP_TCP_H
+#define NNG_TRANSPORT_TCP_TCP_H
+
+#include <nng/nng.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// TCP transport.  This is used for communication over TCP/IP.
```
gdamore, line 20:
```suggestion
// File descriptor transport.  This is used for communication over application supplied file descriptors.
```

include/nng/transport/file_descriptor/file_descriptor.h
```diff
@@ -0,0 +1,30 @@
...
+
+#ifndef NNG_ELIDE_DEPRECATED
+NNG_DECL int nng_tcp_register(void);
```
gdamore, line 23: Delete these lines.

include/nng/transport/file_descriptor/file_descriptor.h
```diff
@@ -0,0 +1,30 @@
...
+// Copyright 2021 Staysail Systems, Inc. <info@staysail.tech>
+// Copyright 2017 Capitar IT Group BV <info@capitar.com>
+//
+// This software is supplied under the terms of the MIT License, a
+// copy of which should be located in the distribution where this
+// file was obtained (LICENSE.txt).  A copy of the license may also be
+// found online at https://opensource.org/licenses/MIT.
+//
+
+#ifndef NNG_TRANSPORT_TCP_TCP_H
```
gdamore, line 11:
```suggestion
#ifndef NNG_TRANSPORT_FILE_DESCRIPTOR_FILE_DESCRIPTOR_H
```

This seems really long -- I almost think we should use the abbreviation "fd" in most places (not the URL though, as that needs to be sure not to collide with other URL schemes that we don't control.)

src/core/file_descriptor.c
```diff
@@ -0,0 +1,455 @@
...
+		nni_aio_finish_error(aio, rv);
+
+		if (nni_list_empty(&d->conaios)) {
+			nni_aio_abort(d->conaio, NNG_ECANCELED);
+			nni_aio_abort(d->resaio, NNG_ECANCELED);
+		}
+	}
+	nni_mtx_unlock(&d->mtx);
+}
+
```
gdamore, line 52: The name resolution parts of TCP should be thrown away-- I forgot about that when we discussed.

src/core/file_descriptor.c
```diff
@@ -0,0 +1,455 @@
...
+file_descriptor_dial_start_next(file_descriptor_dialer *d)
+{
+  /*
+	if (nni_list_empty(&d->conaios)) {
+		return;
+	}
+	nni_resolv_ip(d->host, d->port, d->af, false, &d->sa, d->resaio);
+  */
+}
+
```
gdamore, line 63: In fact, much of this asynchronous dialing part can be thrown away.

Literally dial should just connect to the open FD, and if it returns EBADFD, then we're done. There is no need for an elaborate asynchronous thing -- like there is with TCP.

src/core/file_descriptor.c
```diff
@@ -0,0 +1,455 @@
...
+		nni_aio_finish(aio, 0, 0);
+	}
+
+	file_descriptor_dial_start_next(d);
+	nni_mtx_unlock(&d->mtx);
+}
+
+static void
+file_descriptor_dialer_close(void *arg)
+{
```
gdamore, line 128: When the asynchronous dialer goes away, this function becomes  a no-op.

src/sp/transport/file_descriptor/file_descriptor_test.c
```diff
@@ -0,0 +1,68 @@
...
+
+TEST_LIST = {
+	{ "file descriptor invalid", test_file_descriptor_invalid_fd_fail },
+	{ "file descriptor connect", test_file_descriptor_connect },
+	{ NULL, NULL },
+};
+
+int
+nuts_socket_pair(int *sv)
+{
```
gdamore, line 66: This probably has to be conditionalized since it won't work on UNIX.

src/platform/posix/posix_file_descriptordial.c
```diff
@@ -0,0 +1,416 @@
...
+	}
+	nni_mtx_lock(&d->mtx);
+	d->nodelay = b;
+	nni_mtx_unlock(&d->mtx);
+	return (0);
+}
+
+static int
+file_descriptor_dialer_get_nodelay(void *arg, void *buf, size_t *szp, nni_type t)
+{
```
gdamore, line 283: All these can go away... if the user is giving us a TCP socket, then they are expected to have set this up -- the management of such things will be external to NNG for just plain FDs.

include/nng/nng.h
```diff
@@ -162,24 +167,26 @@ struct nng_sockaddr_storage {
...
+	uint16_t                     s_family;
+	nng_sockaddr_file_descriptor s_file_descriptor;
+	nng_sockaddr_ipc             s_ipc;
+	nng_sockaddr_inproc          s_inproc;
+	nng_sockaddr_in6             s_in6;
+	nng_sockaddr_in              s_in;
+	nng_sockaddr_zt              s_zt;
+	nng_sockaddr_abstract        s_abstract;
+	nng_sockaddr_storage         s_storage;
 } nng_sockaddr;
```
gdamore, line 191: I forgot about the sockaddr stuff... good find.

src/sp/transport/file_descriptor/file_descriptor_test.c
patmaddox: Yeah that was one thing I was wondering. The implementation doesn’t need to be wrapped in a POSIX check, right? Because this is for file descriptors which should exist on any system (I assume… it’s been a long time since I’ve used windows). But generating sockets for the test needs it.

Maybe follow `NNG_HAVE_ABSTRACT_SOCKETS` and do `#ifdef NNG_HAVE_FD_SOCKETS` ?

gdamore:
I'm starting to wonder if we shouldn't just ditch the whole dialing thing altogether - maybe we should allow this to birth a fully formed connection without an endpoint, creating a new API to do so.  I'm not sure whether no having a functional dialer or listener will fit though -- we generally expect that all connected pipes have an endpoint they can point to as a parent.  I want to think about this a bit.

src/sp/transport/file_descriptor/file_descriptor_test.c
gdamore: Yeah, maybe.

I was thinking that maybe this only needs to be done for the test code, but as I think about, I'm not sure file descriptors are universal.  It's a UNIXism, and Windows uses it to some extent as well.  The truly portable thing is FILE *, which is standardized by ANSI C.  Perhaps *that* is the API, although the URL scheme won't work for those.

gdamore:
I've implemented a totally different approach, and merged it.  Please see the nng_socket(7) man page for details.

I'm closing this for now.

PR #1743, opened 2023-12-26
Add a simple SNDPRIO implemention  on req0

### LINKED ISSUE #104: Need RCVPRIO and SNDPRIO options
For legacy compatibility, and perhaps for other use case, it would be nice to support tuning the
send and receive priorities.  The send priority especially would be useful for REQ/REP when
dealing with "less optimal" peers (e.g. over a wider area network).

hoytluo:
fixes #104   Need RCVPRIO and SNDPRIO options

I have added a very simple implemention on req0 to solve my connected two rep server. would you like to merge this into your master?
I add the NNG_OPT_SNDPRIO on the dialer, then copy the sndpri to  the nni_pipe on req0_pipe_start function. then resort the ready_pipes.

Maybe you can  provide better implementation methods. thanks!

### DIFF
--- a/src/sp/protocol/reqrep0/req.c
+++ b/src/sp/protocol/reqrep0/req.c
@@ -76,6 +76,7 @@ struct req0_pipe {
 	bool          closed;
 	nni_aio       aio_send;
 	nni_aio       aio_recv;
+	int           sndprio; 
 };
 
 static void req0_sock_fini(void *);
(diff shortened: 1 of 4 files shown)


src/sp/protocol/reqrep0/req.c
```diff
@@ -190,18 +192,42 @@ req0_pipe_init(void *arg, nni_pipe *pipe, void *s)
...
 {
 	req0_pipe *p = arg;
 	req0_sock *s = p->req;
-
+	int prio = 0;
+	
 	if (nni_pipe_peer(p->pipe) != NNG_REQ0_PEER) {
 		return (NNG_EPROTO);
 	}
-
```
gdamore, line 202: This option should be a generic option, not just on dialers, but on end points, and then configured on the pipe by the end point.

I also think that we should expand this to RECV priority as well, so that we can use the same option for REP to give preference to some peers.  I would like to think on this a bit.

hoytluo: Hi, in the nanomsg code, I find the ep option when I call the nn_connect function, and then copy the socket option(ep_template) to pipe .
But in the nng code, I can't find the ep .  Are the dialer and listener eps? If so, I will add option on the listener also or only add option on the socket, and then copy this option to dialer or listener on creating them.

hoytluo: I have added SNDPRIO to listener also.

PR #1749, opened 2023-12-30
provide NNG_MAX_POLLER_THREADS

gdamore:
This (defaults to 8) sets a limit on the number of poller threads that will be used for servicing I/Os.  This is the size of the I/O completion port thread pool on Windows.  POSIX pollers are generally not concurrent at present.


JaylinYu:
Praise to IOCP

PR #1753, opened 2024-01-01
fix minor documentation details

paddor:


gdamore:
Thanks for fixing these!

PR #1762, opened 2024-01-07
* FIX [core] fix a syntax error

JaylinYu:
Don't know how it bypasses the CI, but clearly a wrong definition in the header file.


gdamore (commented):
It passes CI because its legal.  It is valid to have both named and anonymous parameters. 

Do you have an error message that this fixes?

JaylinYu:
This is reported by one of our customer. 

nevermind, i believe the error is resulted from sth else.

PR #1772, opened 2024-02-13
Simplifies mbedtls CMakeLists.txt

### LINKED ISSUE #1771: The new findMbedTLS from NNG 1.7.2 fails
**Describe the bug**
The improvements to cmake in NNG 1.7.2 prevent MbedTLS being found.

**Actual Behavior**
The final lines when running cmake do not appear for me on Ubuntu 22.04:

```
-- Looking for mbedtls_ssl_init
-- Looking for mbedtls_ssl_init - found
Mbed TLS version: 3.5.2
-- Found MbedTLS: MbedTLS::mbedtls  
-- Configuring done
```
Tested with:
(i) apt installed MbedTLS 2.28  in /usr and
(ii) MbedTLS 3.5.2 installed in /usr/local
Both are not picked up.

This causes the library to be built without TLS support.

**To Reproduce**
However it works again if I simplify this part in `src/supplemental/tls/mbedtls/CMakeLists.txt`:
```
    # If Mbed TLS was added by a consuming project, then we should use that
    # instance of it, instead of configuring our own.
    if (TARGET mbedtls)
        nng_link_libraries(mbedtls)
    else()
        # We want to prefer config mode over our local find package.
        # mbedTLS v3 has a config file, which should work better than
        # what we do here.  We do restore the setting though because
        # user applications might not expect this.
        if (NOT CMAKE_FIND_PAKCAGE_PREFER_CONFIG)
            set(CMAKE_FIND_PACKAGE_PREFER_CONFIG TRUE)
            find_package(MbedTLS REQUIRED)
            set(CMAKE_FIND_PACKAGE_PREFER_CONFIG FALSE)
        else()
            find_package(MbedTLS REQUIRED)
        endif()
        nng_link_libraries_public(MbedTLS::mbedtls MbedTLS::mbedcrypto MbedTLS::mbedx509)
    endif()
```
to this:

```
    # If Mbed TLS was added by a consuming project, then we should use that
    # instance of it, instead of configuring our own.
    if (TARGET mbedtls)
        nng_link_libraries(mbedtls)
    else()
        find_package(MbedTLS REQUIRED)
        nng_link_libraries_public(MbedTLS::mbedtls MbedTLS::mbedcrypto MbedTLS::mbedx509)
    endif()
```
I have no idea what `CMAKE_FIND_PACKAGE_PREFER_CONFIG` is meant for, but perhaps it can be reverted if not important.

Note that there is actually a typo in one of the places where it appears above, but that is not causing this bug.

** Environment Details **
 - NNG version 1.7.2
 - Ubuntu 22.04.3
 - gcc 11.4.0
 - Static lib
 - CMake 3.22.1

shikokuchuo:
Fixes #1771 The new findMbedTLS from NNG 1.7.2 fails

The first commit simplifies the logic such that is works again (as described in the issue).
The second commit just tidies up the new BSD socket transport documentation.

contributing guidelines.

### DIFF
--- a/docs/man/nng_socket.7.adoc
+++ b/docs/man/nng_socket.7.adoc
@@ -16,7 +16,7 @@ nng_socket - BSD Socket transport (experimental)
 
 (((BSD Socket)))(((transport, _socket_)))
 The ((_socket_ transport)) provides communication support between
-peers across a arbitrary BSD sockets, such as those that are
+peers across arbitrary BSD sockets, such as those that are
 created with xref:nng_socket_pair.3supp.adoc[`nng_socket_pair()`].
 
 This transport only supports xref:nng_listener.5.adoc[listeners], using xref:nng_listener_create.3.adoc[`nng_listener_create()`].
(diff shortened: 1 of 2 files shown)


shikokuchuo:
Correcting the typo didn't change anything when I tested. I've updated the issue - something is still not quite right it seems.

gdamore:
This change is wrong.  And I've tested this reasonably myself with a variety of settings.  I think something else is wrong in your environment.  I'm closing this PR for now, because I'm not going to accept it as is.  I think more investigation of the problem is required first.

PR #1810, opened 2024-04-15
tls: reap the connection mutex

gdamore:
On some systems mutexes acquire memory, and we need to make sure finalize the mutex.

This was reported by Jaylin Yu.

contributing guidelines.


gdamore:
Closing in favor of #1811

PR #1824, opened 2024-04-24
macOS - fix deadlock on reqrep socket close

### LINKED ISSUE #1813: Deadlock during nng_close() - multi platform
**Describe the bug**

Continued from the closed https://github.com/nanomsg/nng/issues/1543

When calling `nng_close()` there is sometimes a deadlock which causes `nng_close()` to hang.
This happens also when using only sync APIs (no AIOs).

**Expected behavior**

`nng_close` should finish successfully.

**Actual Behavior**

`nng_close()` hangs.

**To Reproduce**

I created a modified version of the `reqrep` example code here (I use it with IPC transport): [https://gist.github.com/mikisch81/428c4ad87afcc1c8881b282cd5e80eb3](https://gist.github.com/mikisch81/428c4ad87afcc1c8881b282cd5e80eb3)

In the modified example in the client code right before calling `nng_recv` for the reply from the server I start a thread which just calls `nng_close`, after a couple of successful runs the deadlock happen:

```
mischwar@tlv-mpawy reqrep % ./reqrep client ipc:///tmp/reqrep_test
1712848018 - CLIENT: SENDING DATE REQUEST
1712848018 - CLIENT: WAITING FOR REPLY
1712848018 - CLIENT: CLOSING SOCKET
nng_recv error - Object closed
...
...
1712848018 - CLIENT: SENDING DATE REQUEST
1712848018 - CLIENT: WAITING FOR REPLY
1712848018 - CLIENT: CLOSING SOCKET
nng_recv error - Object closed
1712848018 - CLIENT: SENDING DATE REQUEST
1712848018 - CLIENT: WAITING FOR REPLY
1712848018 - CLIENT: CLOSING SOCKET. <--- The thread is suck in nng_close here
```

**Environment Details**

 - NNG version: 1.7.3
 - Operating system and version: MacOS Sonoma 14.4.1 (but also happens on Windows and Linux)
 - Compiler and language used: C/C++ clang
 - Shared or static library - static

**Additional context**

Here is a snapshot of the threads in the client app during the deadlock reproduction:
Screenshot 2024-04-11 at 5 32 52 PM

I recall that the initial suspect was an application callback which is not done:

> Being blocked on nni_aio_wait (as above) and especially task_wait is in indication that your application has a callback executing that is not completing.

So in this example code there is no application callback at all and only blocking APIs are called.
The expected result (as I understand) is that `nng_recv` will always return ECLOSED.

alzix:
when an `aio` has no `a_cancel_fn` and the task is in `task_prep`state abort it on `nni_aio_stop` call

fixes #1813 Deadlock during nng_close() - multi platform

### DIFF
--- a/src/core/aio.c
+++ b/src/core/aio.c
@@ -136,6 +136,8 @@ nni_aio_fini(nni_aio *aio)
 
 	if (fn != NULL) {
 		fn(aio, arg, NNG_ECLOSED);
+	} else {
+		nni_task_abort(&aio->a_task);
 	}
 
 	nni_task_fini(&aio->a_task);
@@ -215,6 +217,8 @@ nni_aio_stop(nni_aio *aio)
 
 		if (fn != NULL) {
 			fn(aio, arg, NNG_ECANCELED);
+		} else {
+			nni_task_abort(&aio->a_task);
 		}
 
 		nni_aio_wait(aio);
@@ -240,6 +244,8 @@ nni_aio_close(nni_aio *aio)
 
 		if (fn != NULL) {
 			fn(aio, arg, NNG_ECLOSED);
+		} else {
+			nni_task_abort(&aio->a_task);
 		}
 	}
 }
@@ -439,6 +445,8 @@ nni_aio_abort(nni_aio *aio, int rv)
 	// Stop any I/O at the provider level.
 	if (fn != NULL) {
 		fn(aio, arg, rv);
+	} else {
+		nni_task_abort(&aio->a_task);
 	}
 }
 


gdamore:
This looks like a good find.  Furthermore, it looks like nni_aio_abort also suffers from the same flaw.

I want to look at this in more detail later today before I move forward.

gdamore:
I think I've convinced myself that this is precisely the right fix, and we just need to add the same change to the implementation of nni_aio_abort.

alzix:
i'm on it

gdamore:
So there are other callers.

Basically we also need the same logic in nni_aio_cancel, and I think nni_aio_fini.  It looks like it was missed in all the paths where we tear down or abort an aio.

gdamore (approved):
This was an excellent find, and probably tricky to do so as well.  Thank you for your contribution!

alzix:
it does not solve the Windows issue though :(
On Windows the issue seems to be different
image

it seems that on Windows it is a TOCTOU issue in the `nni_sock_shutdown`

it is stuck on 
```
// We have to wait for pipes to be removed.
while (!nni_list_empty(&sock->s_pipes)) {
=>	nni_cv_wait(&sock->s_cv);
}
```
while `s_pipes` is already empty
image

the issue predates the fix - so it is better to address it in another PR

gdamore:
I'm merging this... the hang waiting for pipes to be empty feels like a missed cv_wake somewhere.  I'll look for it later. I'm out of time for today.

alzix:
created a new ticket for this #1827

gdamore:
Coming back to this ... I'm now thinking that this change is responsible for a use-after-free.

Essentially, we cannot simply "abort" the task, if there is no cancellation then the task has to run to completion.

I'm now quite interesting to understand what the original hang was that this was supposed to fix.

gdamore:
Ah, I think I see one possible problem.

If we call nni_aio_abort() between a prep, and a schedule, that could lead to problems.
Its better to leave a flag like we do for nni_aio_stop / nni_aio_close so that schedule can see that the operation has been canceled between the begin and the schedule call.

Then schedule can return an error properly.

Issue #1837, opened 2024-06-02
IPC - Use After Free


itayzafrir:
**Describe the bug**
**_Use after free_** in `nni_list_node_remove` leads to process crash.
This is a continuation of https://github.com/nanomsg/nng/issues/1827 and https://github.com/nanomsg/nng/pull/1831 which were closed.
https://github.com/nanomsg/nng/pull/1834 did not fix the use after free issue - at least for the IPC.

**Expected behavior**
No user after free, process does not crash.

**Actual Behavior**
`ipc_send_cb` is called after the pipe has been already closed, causing use after free crash.

**To Reproduce**
Use branch from https://github.com/itayzafrir/nng/tree/itay/new-use-after-free and run the `demo/reqrep.c` executable as server and another one as client.

** Environment Details **
 - NNG version: master branch
 - Operating system and version: Windows 10
 - Compiler and language used: C/C++ Clang
 - Shared or static library: static

**Additional context**
Stack Trace of crash (same as before):
<img width="809" alt="image" src="https://github.com/nanomsg/nng/assets/30868205/626dcb17-f266-4c00-8489-a771ba2f233d">

<img width="815" alt="image" src="https://github.com/nanomsg/nng/assets/30868205/0467eaef-6c67-45f6-b400-85e48661f0be">

Looks like `aio` has already been freed/deleted.

Another point, sometimes the client process crashes with a segfault. I wasn't able to debug this yet but it does happen from time to time.

gdamore:
There is still indeed a use after free here.  I spent pretty much most of last weekend trying to debug it.  It seems like our attempt to unregister the method from the callbacks is not quite complete, or Windows is issuing a second completion callback.

We should not have had any outstanding requests when this callback was fired, as we try to ensure that there are no such requests pending or likely to be queued when we call close.

I'll have another look in the next day or two -- maybe something will occur.

alzix:
please see this post https://qualapps.blogspot.com/2010/05/understanding-readdirectorychangesw_19.html
the Shutting down section. 
it seems the `CancelIo` foes not reliably aborts all future events...
> I re-read the documentation for CancelIo, which makes the statement that "All I/O operations that are canceled complete with the error ERROR_OPERATION_ABORTED, and all completion notifications for the I/O operations occur normally." Decoded, this means that all Completion Routines will be called at least one final time after CancelIo is called.

perhaps we could add `HasOverlappedIoCompleted` busy wait loop.

gdamore:
[...]

PR #1848, opened 2024-07-21
trial to fix winipc crash

gdamore:

contributing guidelines.

### DIFF
--- a/src/platform/windows/win_ipcconn.c
+++ b/src/platform/windows/win_ipcconn.c
@@ -30,6 +30,8 @@ typedef struct ipc_conn {
 	int           send_rv;
 	int           conn_rv;
 	bool          closed;
+	bool          sending;
+	bool          recving;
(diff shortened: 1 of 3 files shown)


gdamore:
This is just a trial balloon.

gdamore:
This PR is an attempt to fix #1837 -- I don't know if it will actually do the trick.  But it passed at least the first round of tests.

itayzafrir:
Hi @gdamore I tested this change on a Windows (virtual) machine & very happy to say that the use after free did not reproduce. Looks like you fixed it 👍

gdamore:
Not precisely as is.  But something close to it

• Garrett

gdamore:
This had an WSS failure on close, which makes me thing we haven't really found them all.  I am thinking I need to take a similar approach for the connect() and accept() calls.  I'd appreciate any further results from folks testing.

PR #1867, opened 2024-09-05
fixes #1866 Use clock_gettime on Android

### LINKED ISSUE #1866: `clock_gettime` is not used on Android, even though available (1.6.0+)
**Describe the bug**
Android has its own implementation of the system libraries: https://android.googlesource.com/platform/bionic.
`clock_gettime` is made available in `libc`, not in `librt` like on some other POSIX platforms. Yet, the check in the build system assumes `clock_gettime` is in `librt`:

From `src/platform/posix/CMakeLists.txt`:
```
nng_check_lib(rt clock_gettime NNG_HAVE_CLOCK_GETTIME)
```

As a consequence, on Android `NNG_HAVE_CLOCK_GETTIME` is never set, and fallback to `nni_clock` is used instead of `clock_gettime`. As documented in the file, this severely hinders stability of nng on that platform

**To Reproduce**
Build for Android, assert in `posix_clock.c` on which implementation is used. Expectation: `clock_gettime`.

**Environment Details**
 - NNG version: started in 1.6.0, with this change: https://github.com/nanomsg/nng/issues/1576
 - Operating system and version: Android
 - Compiler and language used: NDK 26.1.10909125
 - Shared or static library: should not matter

hungrycat:
fixes #1866 `clock_gettime is not used on Android, even though available (1.6.0+)`

On Android's bionic implementation of the system libraries, clock_gettime is available in libc.

### DIFF
--- a/src/platform/posix/CMakeLists.txt
+++ b/src/platform/posix/CMakeLists.txt
@@ -29,7 +29,13 @@ if (NNG_PLATFORM_POSIX)
     nng_check_func(getrandom NNG_HAVE_GETRANDOM)
     nng_check_func(arc4random_buf NNG_HAVE_ARC4RANDOM)
 
-    nng_check_lib(rt clock_gettime NNG_HAVE_CLOCK_GETTIME)
+    if (CMAKE_SYSTEM_NAME MATCHES "Android")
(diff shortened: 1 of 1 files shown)


gdamore:
Oh I didn't see this PR... sorry, I think I fixed it slightly differently.

PR #1870, opened 2024-09-07
fixes #1866 clock_gettime is not used on Android, even though available

### LINKED ISSUE #1866: `clock_gettime` is not used on Android, even though available (1.6.0+)

gdamore:

contributing guidelines.


hungrycat:
Thanks for picking it up. I tested your patch in our project, works as expected. Thank you.

PR #1874, opened 2024-09-25
Update url.c

MuNianCi:
When the cross-compiled versions of the protocol stack and the application are inconsistent, the raw URL may sometimes be cleared, leading to a core dump. 

aarch64--glibc--stable-2022.03-1
gcc-linaro-6.5.0-2018.12-x86_64_aarch64-linux-gnu

### DIFF
--- a/src/core/url.c
+++ b/src/core/url.c
@@ -294,6 +294,9 @@ nni_url_default_port(const char *scheme)
 int
 nni_url_parse(nni_url **urlp, const char *raw)
 {
+	if(NULL == raw){
+		return(NNG_EINVAL);
+	}
 	nni_url *   url;
 	size_t      len;
 	const char *s;


gdamore:
This makes no sense.  You've passed in garbage (a NULL pointer is garbage here), this is an application (caller) error.

Additionally the code is not consistent with the rest of the style.  (We use spaces different -- please use clang format, and we always compare variable's against NULL like :

````
if (value == NULL) {
```

(NULL on the right side).

But again, this PR seems to be fixing a non-issue -- or at least an issue that does not exist with the library.

You'll find e.g. libc doesn't check parameters to strcpy() or strlen() this way.  The assumption is that if you're passing in a URL string here, it's valid.  If the input pointers are junk, then you get what you deserve. :-)

gdamore:
I'm almost certainly going to close this without accepting it.  I'm leaving it open to give the submitter a chance to reply however.

gdamore:
Closing.

PR #1877, opened 2024-10-01
Fixes Compiler Warnings (Windows)

shikokuchuo:
Resulting from latest changes in #1828 and #1848 affecting the Windows platform only.

Eliminates `-Wint-to-pointer-cast` and `-Wmaybe-uninitialized` warnings. Commits should be self-explanatory.


DoumanAsh:
@gdamore This should be included in 1.9.0 because MSVC fails to compile without it by default (at least for me)

PR #1902, opened 2024-10-30
Allow and demonstrate NNG usage with CMake FetchContent

dksmiffs:
@gdamore, this PR breaks two CONTRIBUTING rules, but for these reasons:

1. I couldn't squash to one commit, because the second commit needs the SHA1 of the first.
2. If you accept this change, a follow-up change will be required to the new demo's `CMakeLists.txt`, to refer to your repo instead of my fork.

I hope you understand this awkwardness was necessary because the change I'm suggesting isn't in your repo yet, and yet my demo's `CMakeLists.txt` requires SHA1 access to the repo itself with the single line suggested change to the top-level `CMakeLists.txt`.

I promise I will respond to review feedback, and not leave this baby on your doorstep.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,7 @@ endif ()
 # to obtain details about how the public library was built, so that we can
 # include or not include code based on what's actually present.
 add_library(nng)
+add_library(nng::nng ALIAS nng)
 
 add_library(nng_testing STATIC EXCLUDE_FROM_ALL)
 target_compile_definitions(nng_testing PUBLIC NNG_STATIC_LIB NNG_TEST_LIB NNG_PRIVATE)
--- a/demo/fetch_content/CMakeLists.txt
+++ b/demo/fetch_content/CMakeLists.txt
@@ -1,0 +1,24 @@
+# This software is supplied under the terms of the MIT License, a
+# copy of which should be located in the distribution where this
+# file was obtained (LICENSE.txt).  A copy of the license may also be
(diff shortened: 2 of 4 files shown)


gdamore:
So lets fix this to refer to the main NNG repo instead.  Let me see if I can take this work and adjust it to work that way.

dksmiffs:
@gdamore, after more thought, I don't like the fact that I've overloaded the meaning of `nng::nng`. The `nng::nng` that I introduced in the top level `CMakeLists.txt` file is a CMake library alias.  That's not the same thing as the value of the `NNG_LIBRARY` variable you have set to `nng::nng` in `cmake/nng-config.cmake.in`.  Maybe those overloaded names wouldn't conflict, but I don't have enough experience with CMake config packages to say for sure.

I would feel much more confident if we renamed the library alias I introduced to something unique like `nng::nnglib`.  If you'd be okay with this tweak, I'd be happy to make the change in this pull request.

dksmiffs:
@gdamore After still more thought, I recommend that you close this pull request without merging it into mainline. Your current way works, and does NOT require elevated privileges. Here are widely used CMake build steps that already meet my original intent for writing this pull request:
```bash
cmake -G Ninja -B ./build -S . -DCMAKE_INSTALL_PREFIX=/some/writeable/directory
cmake --build ./build
cd build
ctest
cmake --build . --target install
```
This is the way I'll be using NNG moving forward. Thanks for your hard work on NNG!

gdamore:
Closed at request of submitter.

PR #1948, opened 2024-11-25
tcp: remove the transport layer reference counting mechanism

gdamore:
The upper SP layer takes care not to destroy and endpoint while it has outstanding pipes.

It's possible that the dueling reference counts between the upper and lower layers have been responsible for some hangs on shutdown sometimes seen with other transports.

contributing guidelines.


gdamore:
This has been overcome by events.  Closing.

PR #1951, opened 2024-11-29
Reference counting changes

gdamore:
This converts the main part of NNG to use reference counting atomics efficiently, instead of some other hacky approaches using locks.   It should be safer, and faster both!


gdamore:
Needs to be redone.  Later.

PR #1952, opened 2024-11-30
Nuts tests

gdamore:
This moves a bunch of tests over to NUTS.


gdamore:
This only adds tests -- the bugs it is uncovering need to be fixed, so merging now.

PR #2017, opened 2024-12-25
nni_aio_defer does not use nni_aio_begin first

gdamore:
This reduces the number of locking calls and reduces contention on the hot eq mutex, by implicitly doing the begin first.

Note that it is vital that nni_aio_begin *not* be called first, and that this function (or nni_aio_begin) be called before doing nni_aio_finish (any variant) to prevent nni_aio_wait returning early.

Also while here, if an operation is stopped with nni_aio_close or nni_aio_stop, it should return NNG_ECLOSED, all the time. (We might later introduce a different error for this.)

contributing guidelines.


gdamore:
OBE.

PR #2026, opened 2024-12-28
tcp: flatten the listener implementation

gdamore:
The endpoints both use a nesting level for some common code and some platform dependent code.  But the common code isn't that much and we have similar patterns for e.g. IPC.

This avoids a layer of indirection in the structs, and extra allocations. The payoff will be even larger for the dialers, but that is next. (Dialers are more complicated because of DNS.)

contributing guidelines.


gdamore:
Windows failures seem to be something newly bizarre in the GitHub instances.  The failure is due to a missing library, which makes no sense whatoever.  Punting on it for now.

PR #2027, opened 2024-12-29
fixes #961 surprising pipe event order

gdamore:

contributing guidelines.


gdamore:
This uses a *single* serialization mutex for *ALL* pipe callbacks.

Needless to say, this can have an impact for scalability of systems that have vast amounts of churn in the the number of connected pipes (which is an antipattern for NNG btw), or where the pipe event callbacks do non-trivial amounts of work.

It is important that pipe event callbacks finish *promptly*.  Otherwise they can have a pretty bad impact on the overall system.  This was true before this logic was added, and it remains true.

PR #2117, opened 2025-04-11
Make pointers to data that is not modified const.

sgh:
This allow to pass constrant string i particular and constant data in general to nng functions.

If this change is accepted I'd like to extend it to other functions before merged.


sgh:
Looking at the main headerfile it seems like the send functions realy are the only functions missing const. Is there a special reason for them not being const?

gdamore:
It was just an oversight.

sgh:
How do I add a "signature" to allow merging?

PR #2119, opened 2025-04-14
* FIX [HTTP] fix a blocking case when freeing http client

JaylinYu:
This is a bug I found in the past 2 days. Easily reproduce with low number of taskq_threads and set small timeout duration to http connect aio.
basically when aio timeout before the connaio of dialer, and users try to free the http_client obj will end in infinite blocking at `nni_http_client_fini`. Possibly at `nni_aio_free(c->aio);` or `nng_stream_dialer_free(c->dialer);` Both racing case is due to the ingnored aio aborting here. Because the aio_begin is called before it is put into the nni_list. I assume you shall abort it no matter if it is in the dialing list.

### DIFF
--- a/src/supplemental/http/http_client.c
+++ b/src/supplemental/http/http_client.c
@@ -168,12 +168,10 @@ http_dial_cancel(nni_aio *aio, void *arg, int rv)
 {
 	nni_http_client *c = arg;
 	nni_mtx_lock(&c->mtx);
+	nni_aio_abort(c->aio, rv);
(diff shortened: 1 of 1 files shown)


gdamore:
Looks like a good fix.

gdamore:
Can you rebase this upon main please?  I always merge fixes to main *first*, then cherry pick down to stable if appropriate.

gdamore:
This also got merged to main, so merging it here as well.

gdamore (approved):

PR #2128, opened 2025-05-17
fixes mbedtls 3.6.3 handshake with NULL server name

elijahr:
An explicit call to `mbedtls_ssl_set_hostname(NULL)` is now required if the hostname should not be verified in handshake. From the mbedtls changelog:

```
= Mbed TLS 3.6.3 branch released 2025-03-24

Default behavior changes
   * In TLS clients, if mbedtls_ssl_set_hostname() has not been called,
     mbedtls_ssl_handshake() now fails with MBEDTLS_ERR_SSL_CERTIFICATE_VERIFICATION_WITHOUT_HOSTNAME if certificate-based authentication of the server is attempted.
     This is because authenticating a server without knowing what name
     to expect is usually insecure. To restore the old behavior, either
     call mbedtls_ssl_set_hostname() with NULL as the hostname, or
     enable the new compile-time option MBEDTLS_SSL_CLI_ALLOW_WEAK_CERTIFICATE_VERIFICATION_WITHOUT_HOSTNAME.
```

### DIFF
--- a/src/supplemental/tls/mbedtls/mbedtls.c
+++ b/src/supplemental/tls/mbedtls/mbedtls.c
@@ -290,9 +290,7 @@ conn_init(nng_tls_engine_conn *ec, void *tls, nng_tls_engine_config *cfg,
 		return (tls_mk_err(rv));
 	}
 
-	if (cfg->server_name != NULL) {
-		mbedtls_ssl_set_hostname(&ec->ctx, cfg->server_name);
-	}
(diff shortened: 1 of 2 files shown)


gdamore:
This needs to be rebased to fix a conflict.  I may just deal with it myself.

gdamore (approved):
Just going to let the tests run, I especially want your new test to run in CI/CD before I merge.

PR #2157, opened 2025-08-29
fixes #2156 Messages may fail to be received when using pubsub pattern for asynchronous API transmission via UDP

### LINKED ISSUE #2156: When the pubsub pattern is used to transmit asynchronous APIs via UDP, messages may sometimes fail to be received
When the pubsub pattern is used to transmit asynchronous APIs via UDP, messages may sometimes fail to be received. The specific behavior is as follows: the message cannot be received at the current time, but when another message is sent next time, the message sent last time can be received. 
Using the following demo, when I send 3 messages from the pub side at a relatively fast speed, the sub side can sometimes only receive 1 message.
Could you please help analyze whether this issue is caused by my improper usage or a problem with the protocol stack? Thank you very much.
// pub.c
`
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>
#include <nng/nng.h>

#define SERVER_URL "udp://192.168.31.3:43017"
#define MESSAGE_COUNT 10

int main() {
    nng_socket sock;
    int ret;

    nng_init(NULL);

    if ((ret = nng_pub0_open(&sock)) != 0) {
        fprintf(stderr, "Failed to open pub socket: %s\n", nng_strerror(ret));
        return 1;
    }

    if ((ret = nng_listen(sock, SERVER_URL, NULL, 0)) != 0) {
        fprintf(stderr, "Failed to listen: %s\n", nng_strerror(ret));
        nng_socket_close(sock);
        return 1;
    }

    printf("Publisher started on %s\n", SERVER_URL);
    printf("Publishing %d messages...\n", MESSAGE_COUNT);

    // Ten messages were sent here, but the sub only received one due to its slow reception speed.
    for (int i = 0; i < MESSAGE_COUNT; i++) {
        char msg[128];
        snprintf(msg, sizeof(msg), "Message %d from publisher", i);  

        if ((ret = nng_send(sock, msg, strlen(msg), 0)) != 0) {
            fprintf(stderr, "Failed to send message: %s\n", nng_strerror(ret));
            break;
        }

        printf("Sent: %s\n", msg);
    }

    nng_socket_close(sock);
    printf("Publisher finished\n");
    printf("wait to finish...\n");

    while (1) {
        sleep(1);
    }

    return 0;
}
`

// sub.c 
`
#include <stdio.h>
#include <string.h>
#include <stdlib.h>
#include <unistd.h>
#include <nng/nng.h>

void recv_cb(void *arg);

typedef struct {
    nng_socket sock;
    nng_aio *aio_recv;
} node_t;

node_t node;

// The sub has only received one message.
void recv_cb(void *arg) {
    printf("app recv msg !!!\n");
    
    if (nng_aio_result(node.aio_recv) != 0) {
        fprintf(stderr, "Receive error: %s\n", nng_strerror(nng_aio_result(node.aio_recv)));
        return;
    }
    
    nng_msg *msg = nng_aio_get_msg(node.aio_recv);
    int len = nng_msg_len(msg);
    char buf[1024] = {0};
    
    memcpy(buf, (char*)nng_msg_body(msg), len);
    
    printf("PUBSUB Aio RecvMsg:");
    for (int i = 0; i < len; i++) {
        printf("%x ", buf[i]);
    }
    printf("\n");
    
    nng_msg_free(msg);
    sleep(5); 
    nng_socket_recv(node.sock, node.aio_recv);
}

int main(int argc, char **argv) {
    if (argc < 2) {
        fprintf(stderr, "%s  <url>\n", argv[0]);
        return 0;
    }
    
    int rv;
    nng_init(NULL);
    
    // init socket
    if ((rv = nng_sub0_open(&node.sock)) != 0) {
        fprintf(stderr, "nng_sub_open: %s\n", nng_strerror(rv));
        return 1;
    }
    
    // create AIO 
    nng_aio_alloc(&node.aio_recv, recv_cb, NULL);
    if (rv != 0) {
        fprintf(stderr, "nng alloc  failed %s\n",  nng_strerror(rv) );
    }
    
    nng_sub0_socket_subscribe(node.sock, "", 0);
    
    rv = nng_dial(node.sock, argv[1], NULL, 0);
    if (rv != 0) {
        fprintf(stderr, "nng dial [%s] failed %s\n", argv[1], nng_strerror(rv) );
    }
    
    // recv
    nng_socket_recv(node.sock, node.aio_recv);
    
    nng_aio_free(node.aio_recv);
    nng_socket_close(node.sock);

    return 0;
}
`

MuNianCi:
If there are messages in the queue, retrieve the messages from the queue during reception without waiting for the udp_recv_data operation to complete.

### DIFF
--- a/src/sp/transport/udp/udp.c
+++ b/src/sp/transport/udp/udp.c
@@ -952,6 +952,7 @@ udp_pipe_recv(void *arg, nni_aio *aio)
 {
 	udp_pipe *p  = arg;
 	udp_ep   *ep = p->ep;
+	nni_msg  *msg;
 
 	nni_aio_reset(aio);
 	nni_mtx_lock(&ep->mtx);
@@ -964,7 +965,17 @@ udp_pipe_recv(void *arg, nni_aio *aio)
 		nni_mtx_unlock(&ep->mtx);
 		return;
 	}
-
+	//If there are messages in the queue, retrieve the messages from the queue during reception 
+	//without waiting for the udp_recv_data operation to complete.
+	if(!nni_lmq_empty(&p->rx_mq)){
+		nni_lmq_get(&p->rx_mq, msg);
+		nni_aio_set_msg(aio, msg);
+		nni_aio_completions_add(
+			&ep->complq, aio, 0, nni_aio_count(aio)
+		);
+		nni_mtx_unlock(&ep->mtx);
+		return;
+	}
 	nni_list_append(&p->rx_aios, aio);
 	nni_mtx_unlock(&ep->mtx);
 }
--- a/src/sp/transport/udp/udp.c
+++ b/src/sp/transport/udp/udp.c
@@ -968,7 +968,7 @@ udp_pipe_recv(void *arg, nni_aio *aio)
 	//If there are messages in the queue, retrieve the messages from the queue during reception 
 	//without waiting for the udp_recv_data operation to complete.
 	if(!nni_lmq_empty(&p->rx_mq)){
-		nni_lmq_get(&p->rx_mq, msg);
+		nni_lmq_get(&p->rx_mq, &msg);
 		nni_aio_set_msg(aio, msg);
 		nni_aio_completions_add(
 			&ep->complq, aio, 0, nni_aio_count(aio)


src/sp/transport/udp/udp.c
```diff
@@ -964,7 +965,17 @@ udp_pipe_recv(void *arg, nni_aio *aio)
 		nni_mtx_unlock(&ep->mtx);
 		return;
 	}
-
+	//If there are messages in the queue, retrieve the messages from the queue during reception 
+	//without waiting for the udp_recv_data operation to complete.
+	if(!nni_lmq_empty(&p->rx_mq)){
+		nni_lmq_get(&p->rx_mq, &msg);
+		nni_aio_set_msg(aio, msg);
+		nni_aio_completions_add(
```
gdamore, line 973: Thank you for working on this.

You shouldn't do this particular step however.

The purpose of the completion queue is to allow us to batch aio completions and run those completions without reacquiring the lock, and should only be done on inbound receive path whereas this is being called downwards.

The right thing to do here is simply finish the aio inline.   You can do that by calling:

``nni_aio_finish(aio, 0, nni_aio_count(aio));` 

You can (and should) do that *after* dropping the lock.  

Also,  it would be good if you would use clang-format to fix the whitespace.  For example `//` is always followed by a space, and we insert spaces elsewhere.  (I will make the style suggestions inline.)

src/sp/transport/udp/udp.c
```diff
@@ -964,7 +965,17 @@ udp_pipe_recv(void *arg, nni_aio *aio)
 		nni_mtx_unlock(&ep->mtx);
 		return;
 	}
-
+	//If there are messages in the queue, retrieve the messages from the queue during reception 
+	//without waiting for the udp_recv_data operation to complete.
+	if(!nni_lmq_empty(&p->rx_mq)){
```
gdamore, line 970:
```suggestion
	if (!nni_lmq_empty(&p->rx_mq)) {
```

clang-format will take care of this for you automatically as this project has a `.clang-format` file.

src/sp/transport/udp/udp.c
```diff
@@ -964,7 +965,17 @@ udp_pipe_recv(void *arg, nni_aio *aio)
 		nni_mtx_unlock(&ep->mtx);
 		return;
 	}
-
+	//If there are messages in the queue, retrieve the messages from the queue during reception 
+	//without waiting for the udp_recv_data operation to complete.
```
gdamore, line 969:
```suggestion
	// If there are messages in the queue, retrieve the messages from the queue during reception 
	// without waiting for the udp_recv_data operation to complete.
```

src/sp/transport/udp/udp.c
```diff
@@ -964,7 +965,17 @@ udp_pipe_recv(void *arg, nni_aio *aio)
...
+	if(!nni_lmq_empty(&p->rx_mq)){
+		nni_lmq_get(&p->rx_mq, &msg);
+		nni_aio_set_msg(aio, msg);
+		nni_aio_completions_add(
+			&ep->complq, aio, 0, nni_aio_count(aio)
+		);
```
gdamore, line 975: Please move this to the line above... 
In fact, it might work out that this all can go on one line.  clang-format will automatically format for the width.

PR #2205, opened 2025-12-02
Remove diagnostic pragma in posix.c

### LINKED ISSUE #2204: Diagnostic Pragma in posix_pipe.c
There is a GCC pragma for an unused result for `write()`.

https://github.com/nanomsg/nng/blob/7291d128b5c33377283d7a1f2cd13b615f5b0b10/src/platform/posix/posix_pipe.c#L101-L110

Issue is that strict checking systems (e.g. for R) auto-detect the use of diagnostic suppressing pragmas and complain. This one is relatively easy to silence through other means.

shikokuchuo:
Fixes #2204 diagnostic pragma in posix.c

I've been working around this for the R binding [package checks](https://r-pkgs.org/R-CMD-check.html) through other means until now. But thought this fix may be welcome as a way to silence any potential compiler warning without a pragma.

This may or may not accord with your code style, but it doesn't cause any change to the compiled code AFAIU.

## Summary by CodeRabbit

* **Bug Fixes**
  * Eliminated compiler warnings by making return-value handling explicit in low-level platform pipe operations; preserves existing behavior and public interfaces with no user-facing changes.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/platform/posix/posix_pipe.c
+++ b/src/platform/posix/posix_pipe.c
@@ -103,7 +103,7 @@ nni_plat_pipe_raise(int wfd)
 {
 	char c = 1;
 
-	if (write(wfd, &c, 1)) {};
+	if (write(wfd, &c, 1)) {}
 }
 
 void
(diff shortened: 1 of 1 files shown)


gdamore (commented):
Not loving it. Can you explain why this is a problem?

PR #2230, opened 2026-03-13
Add optional runtime swappable custom allocator

schloing:
Ability to override allocation functions with function pointers `nni_malloc_fn`, `nni_calloc_fn` and `nni_free_fn`. These pointers can be updated at runtime using `nng_alloc_set`.

By default, they are set to `malloc`, `calloc` and `free`. Existing NNG API behaviour not impacted.

Functionality disabled by default, and is enabled with CMake option `NNG_ENABLE_CUSTOM_ALLOC`. When `NNG_ENABLE_CUSTOM_ALLOC` is off, 
the function pointers are not used, and `nni_alloc`, `nni_zalloc`, and 
`nni_free` call the corresponding OS functions directly.

## Summary by CodeRabbit

* **New Features**
  * Custom memory allocators: applications can provide malloc/calloc/free implementations via init parameters or a runtime API; these are validated and applied early and used across platforms.

* **Chores**
  * Build configuration updated to include allocator support.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nanomsg/nng/pull/2230?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmake/NNGOptions.cmake
+++ b/cmake/NNGOptions.cmake
@@ -39,8 +39,6 @@ option(NNG_ENABLE_COVERAGE "Enable coverage reporting." OFF)
 # for the public library.
 option(NNG_ELIDE_DEPRECATED "Elide deprecated functionality." OFF)
 
-option(NNG_ENABLE_CUSTOM_ALLOC "Enable overridable function pointers for allocation functions." OFF)
-mark_as_advanced(NNG_ENABLE_CUSTOM_ALLOC)
 
 option(NNG_ENABLE_STATS "Enable statistics." ON)
 mark_as_advanced(NNG_ENABLE_STATS)
--- a/cmake/NNGOptions.cmake
+++ b/cmake/NNGOptions.cmake
@@ -39,6 +39,9 @@ option(NNG_ENABLE_COVERAGE "Enable coverage reporting." OFF)
 # for the public library.
 option(NNG_ELIDE_DEPRECATED "Elide deprecated functionality." OFF)
 
+option(NNG_ENABLE_CUSTOM_ALLOC "Enable overridable function pointers for allocation functions." ON)
+mark_as_advanced(NNG_ENABLE_CUSTOM_ALLOC)
+
 option(NNG_ENABLE_STATS "Enable statistics." ON)
 mark_as_advanced(NNG_ENABLE_STATS)
 
--- a/cmake/NNGOptions.cmake
+++ b/cmake/NNGOptions.cmake
@@ -39,7 +39,7 @@ option(NNG_ENABLE_COVERAGE "Enable coverage reporting." OFF)
 # for the public library.
 option(NNG_ELIDE_DEPRECATED "Elide deprecated functionality." OFF)
 
-option(NNG_ENABLE_CUSTOM_ALLOC "Enable overridable function pointers for allocation functions." ON)
+option(NNG_ENABLE_CUSTOM_ALLOC "Enable overridable function pointers for allocation functions." OFF)
 mark_as_advanced(NNG_ENABLE_CUSTOM_ALLOC)
 
 option(NNG_ENABLE_STATS "Enable statistics." ON)
--- a/cmake/NNGOptions.cmake
+++ b/cmake/NNGOptions.cmake
@@ -40,7 +40,6 @@ option(NNG_ENABLE_COVERAGE "Enable coverage reporting." OFF)
 # for the public library.
 option(NNG_ELIDE_DEPRECATED "Elide deprecated functionality." OFF)
 
-
 option(NNG_ENABLE_STATS "Enable statistics." ON)
 mark_as_advanced(NNG_ENABLE_STATS)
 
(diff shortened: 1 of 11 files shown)


gdamore (changes requested):
There are a few things here:

1. This shouldn't be conditionalized in CMAKE.  We can do it all the time.  Compared to the overhead of actually allocating memory, the function pointer dereference is *nothing* in terms of performance.
2. The allocators should be moved to an nng_init parameter, without a separate function.  Most notably it would be very bad for the application to *change* the allocators during a run -- this could also have *bad* ramifications if you wind up having for example two different libraries, both running on NNG, but doing this separately (thereby stomping on each other.)

Originally I assumed that changing the allocators would be done as part of a platform port, but I can see other uses here, so I'm supportive of this change in general.  Thanks for contributing!

gdamore:
Needs a rebase.

src/nng.c
```diff
@@ -8,6 +8,7 @@
 // found online at https://opensource.org/licenses/MIT.
 //
 
+#include "core/defs.h"
```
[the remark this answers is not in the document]
gdamore: Oops!  @schloing please fix!

gdamore (changes requested):
At least one concern (trivial to fix) with missing nng->nni declaration.  Also, please update on latest main.

cmake/NNGOptions.cmake
```diff
@@ -39,6 +39,7 @@ option(NNG_ENABLE_COVERAGE "Enable coverage reporting." OFF)
 # for the public library.
 option(NNG_ELIDE_DEPRECATED "Elide deprecated functionality." OFF)
 
+
```
gdamore: Remove this stray whitespace.

include/nng/nng.h
```diff
@@ -551,6 +551,8 @@ NNG_DECL int nng_ctx_set_int(nng_ctx, const char *, int);
 NNG_DECL int nng_ctx_set_size(nng_ctx, const char *, size_t);
 NNG_DECL int nng_ctx_set_ms(nng_ctx, const char *, nng_duration);
 
+NNG_DECL int nng_alloc_set(void* (*malloc_fn)(size_t), void* (*calloc_fn)(size_t, size_t), void (*free_fn)(void*));
```
gdamore: I think the init params mechanism means we should not need this function.  Lets stick to just the init time setup .

include/nng/nng.h
```diff
@@ -1326,6 +1328,10 @@ typedef struct {
 	// will be used. Default is controlled by NNG_RESOLV_CONCURRENCY
 	// compile time variable.
 	int16_t num_resolver_threads;
+
+	void* (*malloc_fn)(size_t);
+	void* (*calloc_fn)(size_t, size_t);
+	void (*free_fn)(void*);
```
gdamore: One important thing -- in order for free to be generally useful with some advanced allocators (e.g. slab allocators), it is useful to pass the size of the chunk being freed to free; if you look at nng_free() you'll see we do that.

(Allocators that don't need this typically have to store some book keeping data in the block being allocated, which can be less efficient and can also stymie efforts for alignment -- those things are likely to be useful in a future RDMA or shared memory design.)

Also, you'll note that we use "void *"  not "void*" elsewhere in the code base.  Please follow the coding style or format with clang format.

src/core/init.c
```diff
@@ -96,7 +106,9 @@ nng_init(const nng_init_params *params)
 	    ? params->num_resolver_threads
 	    : NNG_RESOLV_CONCURRENCY;
 
-	if (((rv = nni_plat_init(&init_params)) != 0) ||
+	if (
+		((rv = nni_alloc_set(init_params.malloc_fn, init_params.calloc_fn, init_params.free_fn)) != 0) ||
+		((rv = nni_plat_init(&init_params)) != 0) ||
```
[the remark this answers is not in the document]
gdamore: This needs to be fixed, although its an edge case.

src/core/init.c
```diff
@@ -74,6 +75,15 @@ nng_init(const nng_init_params *params)
...
 	}
+	init_params.malloc_fn = params->malloc_fn
+			? params->malloc_fn
+			: malloc;
+	init_params.calloc_fn = params->calloc_fn
+			? params->calloc_fn
+			: calloc;
+	init_params.free_fn = params->free_fn
+			? params->free_fn
+			: free;
```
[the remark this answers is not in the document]
gdamore: Absolutely agree here.

src/platform/posix/posix_alloc.c
```diff
@@ -17,20 +17,20 @@
...
-	return (sz > 0 ? calloc(1, sz) : NULL);
+	return (sz > 0 ? nni_calloc_fn(1, sz) : NULL);
 }
 
 void
 nni_free(void *ptr, size_t size)
 {
 	NNI_ARG_UNUSED(size);
-	free(ptr);
+	nni_free_fn(ptr);
```
gdamore: You can see here that nni_free() takes the size.  This is not by accident or oversight.

(For example, in a Solaris kernel, the kernel memory allocator requires the size to be passed to kmem_free().)

gdamore (changes requested):

gdamore (approved):

PR #2243, opened 2026-04-13
fix: `sendmsg`/`readv` EINVAL for large iov on macOS and Windows

### LINKED ISSUE #2242: `sendmsg`/`readv` EINVAL for large iov on macOS and Windows
**Describe the bug**

When testing in downstream R packages which use nanonext, the R binding: attempting to send doubles vectors larger than ~2GiB would fail and cause the pipe to be dropped.                                                            

**To Reproduce**

Claude pinpointed the bug and produced the following self-contained C test repro:
https://gist.github.com/shikokuchuo/f68581d894af32892dc4878cc82cb850
(this demonstrates the issue and doesn't involve any NNG or wrapper code) 

** Environment Details **
 - NNG version: 1.11
 - Operating system and version MacOS: arm64-apple-darwin25.4.0
 - Compiler and language used: clang 21.0
 - Shared or static library: static

PR to follow.

shikokuchuo:
Fixes #2242 <`sendmsg`/`readv` EINVAL for large iov on macOS and Windows>

I've adopted this fix in the current dev version of https://github.com/r-lib/nanonext/pull/267 and I can confirm it fixes the issue.

--

On macOS (XNU kernel), `sendmsg()` and `readv()` return `EINVAL` when the total bytes across all iov entries exceeds `INT_MAX` (2,147,483,647). The XNU socket layer uses 32-bit signed integers internally for aggregate iov size calculations.

NNG's POSIX transport layer passes iov lengths directly from the SP transport without capping. For payloads larger than ~2 GiB, the body `iov_len` exceeds `INT_MAX`, and the syscall fails.

The failure is silent from the caller's perspective: `nng_sendmsg()` returns 0 because the pair protocol completes the application AIO immediately upon accepting the message. The `EINVAL` occurs asynchronously in the I/O thread, which closes the pipe. The receiver then either times out or gets a connection error.

On Windows, a similar truncation issue exists: `WSABUF.len` is `ULONG` (32-bit), so casting from 64-bit `size_t` silently truncates for large buffers.

Capping total iov bytes to `INT_MAX` per syscall resolves this. The SP transport already handles partial I/O, so the cap is transparent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevent integer overflow when aggregating scatter/gather I/O by clamping per-buffer lengths during iovec/WSABUF construction across platforms, improving stability for very large multi-buffer sends/receives.

* **Tests**
  * Added IPC and TCP tests that exercise oversized I/O vectors to verify transfers succeed and reported byte counts do not exceed platform limits.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nanomsg/nng/pull/2243?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/core/aio.c
+++ b/src/core/aio.c
@@ -14,6 +14,7 @@
 #include "reap.h"
 #include "taskq.h"
 
+#include <limits.h>
 #include <string.h>
 
 struct nni_aio_expire_q {
@@ -739,6 +740,13 @@ nni_aio_iov_count(nni_aio *aio)
 		residual += aio->a_iov[i].iov_len;
 	}
 	return (residual);
+}
+
+size_t
+nni_aio_iov_clamp_len(size_t len, size_t count)
+{
+	size_t headroom = (size_t) INT_MAX - count;
+	return (len < headroom ? len : headroom);
 }
 
 size_t
--- a/src/core/aio.c
+++ b/src/core/aio.c
@@ -742,11 +742,16 @@ nni_aio_iov_count(nni_aio *aio)
 	return (residual);
 }
 
-size_t
-nni_aio_iov_clamp_len(size_t len, size_t count)
+bool
+nni_aio_iov_clamp_len(size_t *len, size_t *count)
 {
-	size_t headroom = (size_t) INT_MAX - count;
-	return (len < headroom ? len : headroom);
+	size_t headroom = (size_t) INT_MAX - *count;
+	bool   clamped  = *len > headroom;
+	if (clamped) {
+		*len = headroom;
+	}
+	*count += *len;
+	return clamped;
 }
 
 size_t
--- a/src/core/aio.c
+++ b/src/core/aio.c
@@ -745,6 +745,7 @@ nni_aio_iov_count(nni_aio *aio)
 bool
 nni_aio_iov_clamp_len(size_t *len, size_t *count)
 {
+	NNI_ASSERT(*count <= (size_t) INT_MAX);
 	size_t headroom = (size_t) INT_MAX - *count;
 	bool   clamped  = *len > headroom;
 	if (clamped) {
--- a/src/platform/posix/posix_ipcconn.c
+++ b/src/platform/posix/posix_ipcconn.c
@@ -57,7 +57,7 @@ ipc_dowrite(ipc_conn *c)
 		for (niov = 0, i = 0; i < naiov; i++) {
 			if (aiov[i].iov_len > 0) {
 				size_t len = aiov[i].iov_len;
-				if (count + len > INT_MAX)
+				if (len > INT_MAX - count)
 					len = INT_MAX - count;
 				iovec[niov].iov_len  = len;
 				iovec[niov].iov_base = aiov[i].iov_buf;
@@ -132,7 +132,7 @@ ipc_doread(ipc_conn *c)
 		for (niov = 0, i = 0; i < naiov; i++) {
 			if (aiov[i].iov_len != 0) {
 				size_t len = aiov[i].iov_len;
-				if (count + len > INT_MAX)
+				if (len > INT_MAX - count)
 					len = INT_MAX - count;
 				iovec[niov].iov_len  = len;
 				iovec[niov].iov_base = aiov[i].iov_buf;
--- a/src/platform/posix/posix_ipcconn.c
+++ b/src/platform/posix/posix_ipcconn.c
@@ -13,6 +13,7 @@
 #include "platform/posix/posix_peerid.h"
 #include <errno.h>
 #include <fcntl.h>
+#include <limits.h>
 #include <poll.h>
 #include <stdbool.h>
 #include <string.h>
@@ -52,10 +53,15 @@ ipc_dowrite(ipc_conn *c)
 		int           niov;
 		unsigned      i;
 
+		size_t count = 0;
 		for (niov = 0, i = 0; i < naiov; i++) {
 			if (aiov[i].iov_len > 0) {
-				iovec[niov].iov_len  = aiov[i].iov_len;
+				size_t len = aiov[i].iov_len;
+				if (count + len > INT_MAX)
+					len = INT_MAX - count;
+				iovec[niov].iov_len  = len;
 				iovec[niov].iov_base = aiov[i].iov_buf;
+				count += len;
 				niov++;
 			}
 		}
@@ -122,10 +128,15 @@ ipc_doread(ipc_conn *c)
 			nni_aio_finish_error(aio, NNG_EINVAL);
 			continue;
 		}
+		size_t count = 0;
 		for (niov = 0, i = 0; i < naiov; i++) {
 			if (aiov[i].iov_len != 0) {
-				iovec[niov].iov_len  = aiov[i].iov_len;
+				size_t len = aiov[i].iov_len;
+				if (count + len > INT_MAX)
+					len = INT_MAX - count;
+				iovec[niov].iov_len  = len;
 				iovec[niov].iov_base = aiov[i].iov_buf;
+				count += len;
 				niov++;
 			}
 		}
--- a/src/platform/posix/posix_ipcconn.c
+++ b/src/platform/posix/posix_ipcconn.c
@@ -13,7 +13,6 @@
 #include "platform/posix/posix_peerid.h"
 #include <errno.h>
 #include <fcntl.h>
-#include <limits.h>
 #include <poll.h>
 #include <stdbool.h>
 #include <string.h>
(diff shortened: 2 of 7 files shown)


src/platform/posix/posix_ipcconn.c
```diff
@@ -52,10 +53,15 @@ ipc_dowrite(ipc_conn *c)
 		int           niov;
 		unsigned      i;
 
+		size_t count = 0;
 		for (niov = 0, i = 0; i < naiov; i++) {
 			if (aiov[i].iov_len > 0) {
-				iovec[niov].iov_len  = aiov[i].iov_len;
+				size_t len = aiov[i].iov_len;
+				if (len > INT_MAX - count)
```
gdamore: I'm not sure INT_MAX is well defined everywhere!  Perhaps this should be inside an #ifdef INT_MAX block?

src/platform/posix/posix_ipcconn.c
```diff
@@ -122,10 +128,15 @@ ipc_doread(ipc_conn *c)
 			nni_aio_finish_error(aio, NNG_EINVAL);
 			continue;
 		}
+		size_t count = 0;
 		for (niov = 0, i = 0; i < naiov; i++) {
 			if (aiov[i].iov_len != 0) {
-				iovec[niov].iov_len  = aiov[i].iov_len;
+				size_t len = aiov[i].iov_len;
```
gdamore: Same.

gdamore: I think there's actually a serious bug here.  You are limiting the length of an individual scatter/gather element, but there can be additional ones... if there are additional ones after it, we wind up *skipping* the end of the first oversize element, then adding the next one. 

Also, is the problem here with the size of an individual scatter/gather element, or with the total size? If its the total size (which I suspect may be the case), maybe we should keep a global limit.  That would also address the size of individual elements and solve the above bug.

gdamore (changes requested):
I think there are bugs here.  INT_MAX should be guarded by a define if we're going to use it.  (Or alternatively if we just want to have our own value of it declare "max_transfer" or something as 1<<31.)

This is sticky enough that I'd *really* like to see test coverage for this as well.

src/platform/posix/posix_ipcconn.c
shikokuchuo: Yes, the issue is total size and the clamp is cumulative, not per-element. `count` is a running total; each entry is capped at `INT_MAX - count`. Once the total hits `INT_MAX`, every later entry clamps to 0 — so we always get a contiguous prefix of the original `aiov[]`, with no gaps or reordering.

No bytes are dropped either. Every SP transport send/recv callback (in `src/sp/transport/*`) does:

```c
n = nni_aio_count(aio);
nni_aio_iov_advance(aio, n);
if (nni_aio_iov_count(aio) > 0) {
    // re-dispatch with the remainder
}
```

So a 3 GiB + 1 GiB transfer turns into two syscalls instead of one: the first carries up to `INT_MAX`, `nni_aio_iov_advance` consumes exactly that much of the original `aiov[]`, and the re-dispatch carries the rest (which now fits under the cap).

I've pushed a refactor in [7947d75](https://github.com/nanomsg/nng/pull/2243/commits/7947d75f9b075ee7e8bd611c0fc2608b3a2b4403) that extracts the clamp into `nni_aio_iov_clamp_len()` so the cumulative nature reads more easily at each call site.

src/platform/posix/posix_ipcconn.c
shikokuchuo: In [7947d75](https://github.com/nanomsg/nng/pull/2243/commits/7947d75f9b075ee7e8bd611c0fc2608b3a2b4403) I've refactored so `INT_MAX` is used in exactly one place in `aio.c` and for that file we include `limits.h`, which has been a standard header since C89.

shikokuchuo:
> This is sticky enough that I'd really like to see test coverage for this as well.

A real test would involve sending >2GB of data, which isn't exactly CI-friendly - FWIW I have tested with nanonext and downstream users of nanonext who do use this with large data.

gdamore:
> > This is sticky enough that I'd really like to see test coverage for this as well.
> 
> A real test would involve sending >2GB of data, which isn't exactly CI-friendly - FWIW I have tested with nanonext and downstream users of nanonext who do use this with large data.

It should be possible to fabricate a "send" with > 2GB from memory perhaps using an IOV to resend the same data. :-) 

I think our CI/CD can handle content of this size without too much difficulty.

src/core/aio.c
```diff
@@ -741,6 +742,13 @@ nni_aio_iov_count(nni_aio *aio)
 	return (residual);
 }
 
+size_t
```
gdamore, line 745: This certainly deserves a comment.  Why are we clamping this limit, and why to 2^31?  The answer is that some OS' cannot move more data than this -- while this is clear to us right now, it might not be clear in 5 years time when either I or someone else is looking at this in the future.

gdamore: Actually given that this is a platform limitation, I'm not sure that aio is the place to do it.

One interesting consideration is that some platforms may place limits on the S/G list length, and now we have a different problem, and I'm not sure that all possible consumers here have a ready retry mechanism to do partial transfers.  In some cases we might want to hide the details of that transfer underneath in the platform code.

I want to think about this some more.

gdamore: Ah I see, you're only *applying* the clamp in the places that can cope with this.  I feel better about it now.

shikokuchuo:
> It should be possible to fabricate a "send" with > 2GB from memory perhaps using an IOV to resend the same data. :-)

That's a great idea! I've added tests for both IPC and TCP and confirm that they fail on main, but pass on this branch (on MacOS locally).

gdamore:
I think the CodeRabbit suggestion of gating the test on 64-bit is a good one.  While we only really test in these environments right now, someone may want to build and test for an embedded 32-bit platform, and it would be nice if the tests would not fail there.

src/platform/posix/posix_ipcconn.c
```diff
@@ -122,10 +126,14 @@ ipc_doread(ipc_conn *c)
 			nni_aio_finish_error(aio, NNG_EINVAL);
 			continue;
 		}
+		size_t count = 0;
 		for (niov = 0, i = 0; i < naiov; i++) {
 			if (aiov[i].iov_len != 0) {
-				iovec[niov].iov_len  = aiov[i].iov_len;
+				size_t len = nni_aio_iov_clamp_len(
```
gdamore: The design of this loop means that if we hit the limit before we hit the last iov we will wind up with additional iovs at the end, but with zero length.  I'm a little worried that this might exercise an edge case in the platform code.  (It *should* work.)

A better design would be to have clamp_len return a bool indicating whether it clamped, and pass the iov and count by value (it could update the iov in place, in update count as well.)

src/platform/posix/posix_tcpconn.c
```diff
@@ -123,10 +127,14 @@ tcp_doread(nni_tcp_conn *c)
 			nni_aio_finish_error(aio, NNG_EINVAL);
 			continue;
 		}
+		size_t count = 0;
 		for (niov = 0, i = 0; i < naiov; i++) {
 			if (aiov[i].iov_len != 0) {
-				iovec[niov].iov_len  = aiov[i].iov_len;
+				size_t len = nni_aio_iov_clamp_len(
```
gdamore: If you make the suggested change this would actually reduce duplicated code here as well.

gdamore (changes requested):
Looking much better, but I think it can still be improved.

src/platform/posix/posix_ipcconn.c
shikokuchuo: Great, thanks for the review - I've applied your suggestion in [b4c7b2f](https://github.com/nanomsg/nng/pull/2243/commits/b4c7b2fa6768af639a56758112ea5782c978733f).

src/core/aio.c
```diff
@@ -741,6 +742,18 @@ nni_aio_iov_count(nni_aio *aio)
...
+bool
+nni_aio_iov_clamp_len(size_t *len, size_t *count)
+{
+	size_t headroom = (size_t) INT_MAX - *count;
+	bool   clamped  = *len > headroom;
+	if (clamped) {
+		*len = headroom;
+	}
+	*count += *len;
+	return clamped;
```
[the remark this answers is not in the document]
shikokuchuo: This is addressed by the assert in [2fdd57e](https://github.com/nanomsg/nng/pull/2243/commits/2fdd57e4ef61f31086efbd877d8a0ec5fdbd36b8).

gdamore (approved):

PR #2293, opened 2026-07-18
Expose reference counting operations for dialers and listeners

neachdainn:
Currently a partial implementation and a draft just to get a sense of whether or not this is a good addition and a reasonable implementation. If this looks like a reasonable direction, I'll implement it for listeners, add documentation, and add the (coverage) tests.

There are several operations around dialers and listeners that are difficult to use while keeping within Rust's requirement of soundness. For example, `nng_dialer_get_url` returns a pointer that is freed whenever the dialer is closed, which can happen when either the dialer is closed directly (`nng_dialer_close`) or the socket is closed. Because of this, getting a dialer's URL requires that the Rust code guarantee that both objects are alive and in turn largely means that a Rust `Dialer` object has to have ownership of the socket in some form.

By allowing the user a means to keep the objects alive, the safety of those operations can be guaranteed without inverting the ownership model.

### DIFF
--- a/src/core/dialer.c
+++ b/src/core/dialer.c
@@ -219,6 +219,7 @@ nni_dialer_init(nni_dialer *d, nni_sock *s, nni_sp_tran *tran)
 	d->d_closed   = false;
 	d->d_data     = NULL;
 	d->d_ref      = 1;
+	d->d_user_ref = 0;
 	d->d_sock     = s;
 	d->d_tran     = tran;
 	d->d_inirtime = NNI_SECOND / 100; // 10ms
@@ -353,6 +354,22 @@ nni_dialer_hold(nni_dialer *d)
 	return (rv);
 }
 
+int
+nni_dialer_user_hold(nni_dialer *d)
+{
+	int rv;
+	nni_mtx_lock(&dialers_lk);
+	if (d->d_closed) {
+		rv = NNG_ECLOSED;
+	} else {
+		d->d_ref++;
+		d->d_user_ref++;
+		rv = 0;
+	}
+	nni_mtx_unlock(&dialers_lk);
+	return (rv);
+}
+
 void
 nni_dialer_rele(nni_dialer *d)
 {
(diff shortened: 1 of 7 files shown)


src/core/dialer.c
```diff
@@ -353,6 +354,22 @@ nni_dialer_hold(nni_dialer *d)
...
+		rv = NNG_ECLOSED;
+	} else {
+		d->d_ref++;
+		d->d_user_ref++;
+		rv = 0;
+	}
+	nni_mtx_unlock(&dialers_lk);
+	return (rv);
+}
+
```
neachdainn, line 372: Because the `nni_mtx_lock` isn't a recursive lock, we don't have a way to both increment `d_user_ref` and call `nni_dialer_hold` without there being the potential for another thread to free the dialer in the period of time where the lock isn't held. Which is why this is largely a copy-and-paste of that function instead of calling it.

src/core/dialer.c
```diff
@@ -362,13 +379,39 @@ nni_dialer_rele(nni_dialer *d)
...
+		d->d_user_ref--;
+		rele = true;
+		rv   = 0;
+	}
+	nni_mtx_unlock(&dialers_lk);
+
+	if (rele) {
+		nni_dialer_rele(d);
+	}
+	return (rv);
```
neachdainn, line 412: I don't love the fact that this isn't symmetrical with `nni_dialer_user_hold`, in that one is copy-and-paste but the other calls a function. But `hold` can't call a function and the `rele` has enough logic inside of it that I didn't want to risk copy-and-paste bugs, so we get this situation.

src/core/dialer.c
```diff
@@ -362,13 +379,39 @@ nni_dialer_rele(nni_dialer *d)
...
+		nni_id_remove(&dialers, d->d_id);
+	}
 	nni_mtx_unlock(&dialers_lk);
 
 	if (reap) {
 		nni_dialer_reap(d);
 	}
 }
 
+int
```
neachdainn, line 392: I am assuming that "returns `ESTATE` when the user releases too many times" is a good feature. If not, this can be `void` like `nni_dialer_rele`.

gdamore:
Mostly I like it, but I suggested some improvements.  It needs test coverage as well.

neachdainn:
> Mostly I like it, but I suggested some improvements. It needs test coverage as well.

I am very open to suggestions but I'm not seeing any here outside of test coverage. Did I miss them somewhere?

neachdainn:
@gdamore Having dug into it some, I'm questioning  the value of this particular implementation. Without sizable changes in the relationship between the dialer/listener and the socket, it seems that one of things must be true:

1. Closing the socket will block  until all of the user holds are released
2. The dialer/listener needs to hold a handle to the socket instead of a pointer
3. Some operations (e.g., `nng_dialer_get_bool`) become unsafe

The last one means that the user of NNG either has to make certain the socket is always open or track it in some way, which is the very thing I opened this PR to avoid. The second one seems like a hack and will probably have a major impact on performance (I'm assuming - I have not tried). That leaves the blocking close which seems like a suboptimal API.

If you think those aren't the only three options or if you really want this feature, I'll keep digging and working. Otherwise, if I'm going to write bad code, I'd rather put bad code in my library than yours.

PR #2299, opened 2026-07-29
Fix async connect completion failure when dial_aio is set to NULL

MuNianCi:
Setting c->dial_aio to NULL prevents the asynchronous connect completion handler from finding the associated aio, thus the dial cannot be completed or cancelled.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of asynchronous connection handling by preserving connection state during dial operations.

### DIFF
--- a/src/platform/posix/posix_ipcdial.c
+++ b/src/platform/posix/posix_ipcdial.c
@@ -226,7 +226,7 @@ ipc_dialer_dial(void *arg, nni_aio *aio)
 		if ((rv = nni_posix_pfd_arm(&c->pfd, NNI_POLL_OUT)) != 0) {
 			goto error;
 		}
-		c->dial_aio = NULL;
+		// c->dial_aio = NULL;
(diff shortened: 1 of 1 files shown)


src/platform/posix/posix_ipcdial.c
```diff
@@ -226,7 +226,7 @@ ipc_dialer_dial(void *arg, nni_aio *aio)
 		if ((rv = nni_posix_pfd_arm(&c->pfd, NNI_POLL_OUT)) != 0) {
 			goto error;
 		}
-		c->dial_aio = NULL;
+		// c->dial_aio = NULL;
```
gdamore, line 229: I do not believe this is correct.  As you can see below, we move the AIO from this one handle, to the connq, which is where it it will be processed.

I think that not clearing this will lead to other problems.

I need to look at this in more detail -- but I see that the CI/CD also flagged a bunch problems here which may be related.

PR #2304, opened 2026-08-22
fix(security): canonify URLs on the the HTTP server (fixes #2303)

### LINKED ISSUE #2303: Security vulnerability for `nng_http_handler_alloc_directory`
**Describe the bug**

Possibility to traverse the filesystem outside the served folder when using `nng_http_handler_alloc_directory` and
`nng_http_handler_set_tree`.

**Expected behavior**

When setting up an HTTP handler for a directory like `/opt/subdir/folder1` it should not be possible to access any
files/directories other than those that are direct children.

F.i. if file `/opt/subdir/folder2/file.txt` exists, it should not be possible to access it via:

> curl --path-as-is -X GET -s "http://127.0.0.1:8090/../folder2/file.txt"

**Actual Behavior**

Actually it is possible to retrieve non-children files to the served directory.

> curl --path-as-is -X GET -s "http://127.0.0.1:8090/../folder2/file.txt"

yields the contents of the `/opt/subdir/folder2/file.txt`.

Even though "browsing" is not possible, i.e.

> curl --path-as-is -X GET -s "http://127.0.0.1:8090/../folder2/"

will not yield a list of directories and files, if the server is running with root privileges, this 
still constitutes a vulnerability, as f.i. the following would be possible:

> curl --path-as-is -X GET -s "http://127.0.0.1:8090/../../../etc/hostname"
my-host-name

**To Reproduce**

Set up a HTTP server serving any directory using `nng_http_handler_alloc_directory` + `nng_http_handler_set_tree`.

Then use curl with option `--path-as-is` to pass a URI path with relative paths.

**Environment Details**
 - Tested with NNG 1.12
 - All OSs

gdamore:
This fix was backported from the main development branch.

contributing guidelines.

### DIFF
--- a/src/core/url.c
+++ b/src/core/url.c
@@ -116,8 +116,8 @@ nni_url_decode(uint8_t *out, const char *in, size_t max_len)
 	return (len);
 }
 
-static int
-url_canonify_uri(char **outp, const char *in)
+int
+nni_url_canonify_uri(char **outp, const char *in)
 {
 	char *  out;
 	size_t  src, dst, len;
@@ -399,7 +399,7 @@ nni_url_parse(nni_url **urlp, const char *raw)
 	url->u_host[len] = '\0';
 	s += len;
 
-	if ((rv = url_canonify_uri(&url->u_requri, s)) != 0) {
+	if ((rv = nni_url_canonify_uri(&url->u_requri, s)) != 0) {
 		goto error;
 	}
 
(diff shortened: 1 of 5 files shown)


gdamore:
The mbedTLS build on macOS fails because the default changed to mbedTLS 3, and nng 1.0 only supports mbedTLS 2. 

I'll change the workflow for that in a later PR.

PR #2338, opened 2026-09-19
docs: fix the nng1 migration guide's example and API names

NotAFlightRisk:
The abstract socket example under "Abstract Autobinding No Longer Supported" doesn't compile. Its format string opens on a backtick instead of a quote, and the four calls beneath it are `nni_random`, which isnt in the installed headers. The prose one line above already says `nng_random`.

Six other names in the file dont exist in nng2 either - `nni_aio`, `nng_stream_peer_self_addr`, the two HTTP setters spelled `req_set` and `res_set`, and both security descriptor functions spelled `get` rather than `set`. None of the six is in `nng.h` or `http.h`, and none has an `xref.md` entry, so all six print on the published page with the brackets still showing. `NNG_OPT_WS_OPT_WS_REQUEST_HEADERS` has its prefix doubled too.

Five of the corrected names do have `xref.md` entries, so those turn into links. `nng_stream_listener_set_security_descriptor` has no entry under either spelling. That one stays bracketed - I didn't want to invent a page for it.

## Summary by CodeRabbit

- **Documentation**
  - Corrected the NNG 1.x to 2.0 migration guide’s API references.
  - Updated HTTP, WebSocket, stream, asynchronous I/O, security descriptor, and random-number examples to use the correct names and syntax.


gdamore (approved):
Thanks for catching these mistakes!

